Issue #11, opened 2019-11-05
Enforce path options are absolute paths


oliverwm1:
We (myself, @mcgibbon, @nbren12) have agreed (I think) that user-supplied paths to custom files/directories should be specified as absolute paths as a way to distinguish from built-in options. This is not actually implemented right now.

oliverwm1:
Closed by #15

PR #15, opened 2019-11-07
Robustly handle custom/built-in options and allow remote downloads

### LINKED ISSUE #10: Add ability to download files from cloud storage bucket
If user provides a path that begins with gs:// for an option that can be specified as a path to a file or directory (specifically diag_table, data_table, forcing or initial_conditions) then have fv3config download that data and put it in the right place.

I think it's fair to assume the user will have appropriate authorization if pointing to a private storage bucket, but we should ensure fv3config fails gracefully if they do not.

oliverwm1:
Closes #10 and #11 

This PR will improve the handling of custom configuration options. It introduces a function `resolve_option` which distinguishes between custom local paths, custom remote (Google storage bucket) paths and built-in tags. This abstracts away logic that was previously repeated in many places. It also adds functionality to download data from a remote path instead of symlinking from a local path.


oliverwm1:
This PR is ready for review. Note some functions have no unit tests (specifically `link_or_copy_directory`, `copy_file` and `gsutil_is_installed` in _datastore.py) because they depend on `gsutil`. This is about 15 simple lines of code. I have verified that all of these work as intended in my own usage.

tests/test_namelist.py
```diff
@@ -151,6 +175,40 @@ def test_default_config_has_entries(self):
         for name, value in config.items():
             self.assertIsInstance(name, str, f'key {name} is not a string')
 
+    def test_enable_restart_from_default(self):
```
mcgibbon: Could you split these tests off into another TestCase class? They don't use the setup or teardown so it won't break the code, and I think it would help better organize the tests. The only reason I'm asking is for test organization (I find it's nice to have the tests that test one feature in one TestCase). Feel free to disagree.

tests/test_tables.py
```diff
@@ -149,20 +151,22 @@ def test_get_current_date_from_bad_coupler_res(self):
...
-        with open(diag_table_in, 'w') as f:
+        input_directory = os.path.join(rundir, 'INPUT')
+        diag_table_filename = os.path.join(rundir, 'diag_table')
+        with open(diag_table_filename, 'w') as f:
             f.write(diag_table_test_in)
-        write_diag_table(config_for_write_diag_table_test, diag_table_in, diag_table_out)
-        with open(diag_table_out) as f:
+        current_date = get_current_date_from_config(config_for_update_diag_table_test, input_directory)
+        update_diag_table_for_config(config_for_update_diag_table_test, current_date, diag_table_filename)
+        with open(diag_table_filename) as f:
```
mcgibbon: I assume `write_diag_table` got refactored? Just double-checking.

fv3config/_tables.py
```diff
@@ -99,30 +84,33 @@ def get_current_date_from_config(config):
     if force_date_from_namelist:
         current_date = config['namelist']['coupler_nml'].get('current_date', [0, 0, 0, 0, 0, 0])
     else:
-        coupler_res_filename = os.path.join(get_initial_conditions_directory(config), 'coupler.res')
+        coupler_res_filename = os.path.join(input_directory, 'coupler.res')
```
mcgibbon: What was the motivation for passing the input directory separate from config?

fv3config/__init__.py
```diff
@@ -4,7 +4,7 @@
 
 from ._config import (
     config_to_namelist, config_from_namelist, get_default_config,
-    write_run_directory,
+    write_run_directory, enable_restart
```
mcgibbon: Could you add a description to the documentation?

setup.py
```diff
@@ -42,6 +42,7 @@
         ],
     },
     install_requires=requirements,
+    extras_requires={'bucket-access': 'gsutil'},
```
mcgibbon: I didn't know this was an option!

fv3config/_datastore.py
```diff
@@ -166,5 +168,47 @@ def extract_data():
         shutil.move(os.path.join(app_data_dir, filename_root), local_archive_dir)
 
 
+def is_gsbucket_url(path):
+    if path.startswith(gs_bucket_prefix):
```
nbren12: The if statement here isn't necessary. `return path.startswith(gs_bucket_prefix)` is sufficient.

fv3config/_datastore.py
```diff
@@ -166,5 +168,47 @@ def extract_data():
...
+        built_in_options_dict (dict): built-in options
+
+    Returns:
+        (str): a path or url
+
+    Raises:
+        ConfigError: if option is an absolute path but does not exist or if
+                     option is not in default_options_dict
+    """
+    if os.path.isabs(option):
```
nbren12: Perhaps add a comment mentioning that this nested if statement is a reflection of FV3's internal logic, not a design choice on our part.

tests/test_namelist.py
oliverwm1: These tests were introduced in a previous PR (#9) so don't know why they are showing up here. I think this is a GitHub issue. Still, I agree with your point. I created an issue for this and the enable_restart documentation point. See #24

tests/test_tables.py
oliverwm1: Yes, `write_diag_table` was changed to `update_diag_table_for_config` which has slightly different functionality because it changes the diag_table in place instead of writing a new diag_table.

fv3config/_tables.py
oliverwm1: The code needs to check if `coupler.res` exists among the initial conditions files. I didn't want this part of the code to have to deal with downloading from google cloud storage, so the easier option is to pass the local rundir/INPUT, since the initial conditions files have already been linked or copied to that directory when this code is called. Not a perfect solution, but this can be improved if/when we change the data structure for the initial conditions to a class.

fv3config/__init__.py
oliverwm1: Again, `enable_restart` was introduced in #9. But docs are needed. I added an issue #24 .

fv3config/_datastore.py
oliverwm1: Good call. Changed.

fv3config/_datastore.py
oliverwm1: Good idea. Added.

oliverwm1:
> * PR description doesn't include all changes (specifically, `enable_restart`)

As mentioned in some of my specific replies, the `enable_restart` functionality was actually introduced in #9. I don't know why it's showing up here, since this code already exists in master. So I think my PR description is complete. I agree docs/tests could be improved for `enable_restart`, so I added an issue #24.

fv3config/__init__.py
mcgibbon: I also added a Trello card. We should probably get in the habit of putting issues on Trello.

fv3config/_tables.py
mcgibbon: Could you add a space-separated sentence with "The input directory is passed to read date from any initial conditions files, since config can contain a remote data source and we do not want to download data in this function."? You can edit it if you want. This should clarify for any future askers of this question.

tests/test_namelist.py
mcgibbon: Great.

tests/test_tables.py
mcgibbon: Great.

fv3config/_tables.py
oliverwm1: Yup. Added that sentence (with minor edits) to the docstring.

Issue #22, opened 2019-11-13
validate the initial condition data


nbren12:
There are some simple checks we could perform on the initial condition data to make sure it is compatible with the fv3gfs. Here are a couple of examples:

- The restart files need to only contain `float` values, otherwise the model errors out.
- The grid needs to match the one specified in the `input.nml`. 

Ultimately, this feature should probably be part of a bigger refactor of the representation of the initial condition data to a richer python object.

mcgibbon:
![](https://github.trello.services/images/mini-trello-icon.png) [FV3Config: Validate initial conditions data](https://trello.com/c/SP3k0lnk/127-fv3config-validate-initial-conditions-data)

oliverwm1:
Agreed some checking on initial condition data would be useful. This is one aspect of what I suggested in #6 although we got a little sidetracked in that issue.

Issue #6, opened 2019-10-28
Add check_config step in write_run_directory


oliverwm1:
`config.write_run_directory` should check that the config dict provided is valid (to the extent possible). Based on the configuration flags set, it should be able to determine what initial conditions files are needed, and check for the existence of those in the specified initial conditions directory. This would help ensure that user-supplied configurations are still valid.

In theory we could also check for invalid namelist options, but this would be a lot of work to implement, because it involves going through all the source code to pull out default namelists.

PR #27, opened 2019-11-16
Allow per-file specification of config files

### LINKED ISSUE #8: Add "extra_files" option to config dictionary
Allow user to specify extra files that should be symlinked to the rundir. There should be an option to choose what subdirectory within the rundir the files should go.

### LINKED ISSUE #23: Writing to a symlink with xarray.to_netcdf has strange behavior
cc @oliverwm1 

I just figure out what was happening. Saving a dataset with `xarray.to_netcdf` to an existing symlink alters the target of the symlink, rather than removing the symlink and then saving a new file.

Therefore, what was happening was
1. make run directories for "simple" and "complex". these contain symlinks to the same input data files.
2. Patch-in the "simple" surface data (this overwrites the data being pointed to).
3. Patch-in the "complex" surface data 

After step 3,  the target of the symlinks contains the "complex" data, and both the "simple" and "complex" run directories point to this same data.

While this appears to be an xarray issue, maybe we shouldn't be using symlinks for the input data since it opens up the opportunity for these kinds of side-effects. The initial condition data is pretty small, so I don't see much overhead for copying.

### LINKED ISSUE #25: Allow per-file specification of initial conditions and forcing files
The `initial_conditions` and `forcing` options in the `config` dictionary are currently specified as paths to directories. We would like the user to be able to specify the files that make up the initial conditions or forcing files on a per-file basis. 

This could be done by allowing these options to be dicts or lists of tuples which include the source and destination path for each file, as well as information about whether to copy or symlink that particular file.

oliverwm1:
(Edited 2019-11-20 to reflect current PR)

Closes #8, closes #23, closes #25.

In order to allow a per-file specification of initial conditions and forcing files (as well as whether to link or copy particular files) this PR introduces a low-level representation of the `config` dictionary called `asset_list`. This `asset_list` is a list of dicts. Each dict represents one file, and has information about its source name/location, target name and location relative to the root of the run-directory, and whether to copy or link.

API changes:
---------
- no breaking changes
- user can now specify initial conditions or forcing files through a list of dicts
- user can specify a set of patch files through the new, optional, ``config['patch_files']`` (these will overwrite any files specified in initial conditions or forcing)
- no files are symlinked. By default, the orography and forcing files are hard-linked and the initial conditions are copied.

Other notes:
---------
The functions ``link_or_copy_directory``, ``link_directory`` have been removed. ``copy_file`` and ``gsutil_is_installed`` have been moved to the new _asset_list.py. Some test refactoring had to be done because of this.

I put a few of the tests for ``asset_list`` in ``test_forcingdata.py`` because they rely on the existence of downloaded input data and lists of forcing/initial condition data, which already exist in ``test_forcingdata.py``.

Added a missing required file (`fv_core.res.nc`) to the restart initial conditions required filenames in ``test_forcingdata.py``

My method for generating an asset_list from a google cloud storage url is not great (some funky string manipulation on output from `gsutil ls`). We should probably move from using `gsutil` to `gcsfs` but I suggest doing that in a later PR.


oliverwm1:
[Allow per-file specification of run directory in fv3config](https://trello.com/c/fVyKbbmw/138-allow-per-file-specification-of-run-directory-in-fv3config)

mcgibbon:
"We should probably move from using `gsutil` to `gcsfs` but I suggest doing that in a later PR." please open an issue for this, before we forget.

fv3config/_asset_list.py
```diff
@@ -0,0 +1,245 @@
...
+        os.remove(target_item)
+    os.link(source_item, target_item)
+
+
+def copy_file(source_path, target_path):
+    if is_gsbucket_url(source_path):
+        if gsutil_is_installed():
+            check_call(['gsutil', 'cp', source_path, target_path])
+        else:
+            logging.warning(f'Optional dependency gsutil not found. File {source_path} will not be copied to {target_path}')
```
mcgibbon: Please raise an exception instead of a warning. I think this counts as a catastrophic failure.

tests/test_asset_list.py
```diff
@@ -0,0 +1,232 @@
...
+            'source_location': '',
+            'source_name': '',
+            'target_location': '',
+            'target_name': '',
+            'copy_method': '',
+        }
+        try:
+            check_asset_valid(proper_asset_dict)
+        except fv3config.ConfigError:
+            self.fail("check_asset_valid raise ConfigError unexpectedly")
```
mcgibbon: "raise" to "raised"

tests/test_asset_list.py
```diff
@@ -0,0 +1,232 @@
...
+        with self.assertRaises(fv3config.ConfigError):
+            check_asset_valid(bad_asset_dict)
+
+    def test_check_asset_valid_proper_asset(self):
+        proper_asset_dict = {
+            'source_location': '',
+            'source_name': '',
+            'target_location': '',
+            'target_name': '',
+            'copy_method': '',
```
mcgibbon: Is this really a valid asset? copy_method is not one of the valid options, and presumably the names need to be non-empty.

mcgibbon: At some point in the future we may want to use something cool [like `JSON-Schema`](https://stackoverflow.com/questions/3262569/validating-a-yaml-document-in-python) to validate our dictionaries.

docs/usage.rst
```diff
@@ -85,6 +85,31 @@ The ``namelist`` item is special in that it is explicitly stored in the ``config
...
+representation of run-directory files. An asset is a dictionary that knows about one files's source
+location/filename, target filename, target location within the run directory and whether that file is copied or linked.
+Asset dicts can be generated with the helper function ``generate_asset``. For example::
+
+    >>> generate_asset('/path/to/filedir', 'sample_file.nc', target_location='INPUT')
+    {'source_location': '/path/to/filedir',
+    'source_name': 'sample_file.nc',
+    'target_location': 'INPUT',
+    'target_name': 'sample_file.nc',
+    'copy_method': 'copy'}
```
mcgibbon: Should list the possible options in this doc.

docs/usage.rst
```diff
@@ -85,6 +85,31 @@ The ``namelist`` item is special in that it is explicitly stored in the ``config
...
+Specifying individual files
+---------------------------
+
+More fine-grained control of the files that are written to the run-directory is possible using the "asset"
+representation of run-directory files. An asset is a dictionary that knows about one files's source
+location/filename, target filename, target location within the run directory and whether that file is copied or linked.
+Asset dicts can be generated with the helper function ``generate_asset``. For example::
+
+    >>> generate_asset('/path/to/filedir', 'sample_file.nc', target_location='INPUT')
+    {'source_location': '/path/to/filedir',
```
mcgibbon: Should include the trailing "/" to make it slightly clearer at a glance that this is a directory (even if it's already pretty clear from the name).

docs/usage.rst
```diff
@@ -85,6 +85,31 @@ The ``namelist`` item is special in that it is explicitly stored in the ``config
...
+    'source_name': 'sample_file.nc',
+    'target_location': 'INPUT',
```
mcgibbon: Similarly would be ideal to include a trailing "/" to make it clear this is a directory.

docs/usage.rst
mcgibbon: Even though yes, this is already documented in the API autodoc through the docstring.

tests/test_asset_list.py
oliverwm1: Done.

tests/test_asset_list.py
oliverwm1: Changed function name to `check_asset_has_required_keys` to better reflect its purpose.

docs/usage.rst
oliverwm1: Done.

docs/usage.rst
oliverwm1: Done.

docs/usage.rst
oliverwm1: I made ``generate_asset()`` link directly to its definition in the API so that it's easy to find for readers. Do you think that's sufficient? I'd rather not make this usage section much longer (or have the exact same info in two places).

oliverwm1:
> "We should probably move from using `gsutil` to `gcsfs` but I suggest doing that in a later PR." please open an issue for this, before we forget.

Issued opened. #33

fv3config/_asset_list.py
oliverwm1: Done. Also now raise the exception before trying to generate an asset_list from a gcs url if gsutil is not installed.

docs/usage.rst
mcgibbon: Have you tested the link works? I thought it's `:py:meth:`. If it works this is fine.

tests/test_asset_list.py
mcgibbon: I suppose we'll refactor this test later if we implement more stringent checks.

docs/usage.rst
oliverwm1: Yes. The `:py` part is assumed.

tests/test_asset_list.py
oliverwm1: I agree, this could be done as part of implementing #6 and #22

fv3config/_config.py
```diff
@@ -99,21 +92,11 @@ def write_run_directory(config, target_directory):
...
-    initial_conditions_dir = get_initial_conditions_directory(config)
-    base_forcing_dir = get_base_forcing_directory(config)
-    orographic_forcing_dir = get_orographic_forcing_directory(config)
-    field_table_filename = get_field_table_filename(config)
-    diag_table_filename = get_diag_table_filename(config)
-    data_table_filename = get_data_table_filename(config)
+    asset_list = config_to_asset_list(config)
     for directory in [target_directory, input_directory, restart_directory]:
         if not os.path.isdir(directory):
             os.mkdir(directory)
```
nbren12: Can this logic be moved into `write_asset_list`? It seems a little low-level for this top-level routine.

nbren12: In fact, isn't this code unneeded since this line is in `write_asset`?
```
        os.makedirs(os.path.dirname(target_path))
```

oliverwm1: Thanks yeah I just realized this last night looking the code. Used to be needed, but no longer!

fv3config/_asset_list.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def get_field_table_asset(config):
+    """Return asset for field_table"""
+    field_table_filename = get_field_table_filename(config)
+    location, name = os.path.split(field_table_filename)
+    return generate_asset(location, name, target_name='field_table')
+
+
+def generate_asset(source_location, source_name, target_location='',
```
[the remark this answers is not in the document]
oliverwm1: Yeah, I'd be fine with just `asset`. I do think this is different than all the `get_` functions which generate assets from paths to files or directories, whereas `generate_asset` is basically just an initialization function as you say.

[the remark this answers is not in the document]
oliverwm1: Actually this has the disadvantage that it conflicts with using `asset` as a variable name elsewhere (which I do when looping over an `asset_list` for example).

What about `initialize_asset`? I don't think I follow the rationale for why this function doesn't need a verb.

nbren12:
@mcgibbon Can we merge this? It seems like we have both approved it.

Issue #19, opened 2019-11-12
Release managment


ofuhrer:
Here are some thoughts now that fv3config is public...
- Have you thought about release management, now that the repo is public?
- Is now the right time to do a first release tag?
- The documentation talks about a "stable" release using `pip install`. What is installed when somebody does `pip install` (releaseX.X?) and what is installed when somebody installs from source (master?)?

PR #28, opened 2019-11-16
Allow user to specify cache dir through env or function

### LINKED ISSUE #7: Allow specifying download cache location
As brought up by @ofuhrer in VulcanClimateModeling/fv3gfs#41, on production systems we'd like the data cache to go to some kind of scratch folder and not `/home/username`. There should be a way to specify where the download cache should be stored.

I think the "correct" option would be to store the download location in the appdirs folder (`/home/username`) as a string in a text file (yaml?), and allow the user to specify where they want the data to be stored when they call `python -m fv3config.download_data` (e.g. with a `--location=/scratch` command-line argument). When we import the package it would then load the download location from its configuration file.

mcgibbon:
Fixes #7. This PR changes the API in three ways:

- if environment variable `FV3CONFIG_CACHE_DIR` is set, fv3config will use this location for cache when imported (specifically a `fv3config-cache` subdirectory). Will use the user's cache directory if unset (changed from last version where we used the user's application data directory).
- new `fv3config.set_cache_dir(dirname)` can be used to change the cache dir after import
- new `fv3config.get_cache_dir()` will return the current cache directory
- Updated the installation documentation with warnings about this being pre-alpha, helps with #19

Some internal changes that don't affect API:

- data downloading is now tested, using temporary directories and by mocking requests.get
- some global variables have been reformatted to allcaps
- only one global variable in _datastore contains the cache directory, the other ones have been moved to private helper functions that will determine their values based on the current cache directory
- the options dictionaries now map to paths that are relative to the cache directory, and not including that directory

Small bug fix I bundled in this PR:
- fixed the `extras_require` keyword in setup.py, which was being ignored due to a typo


mcgibbon:
- [x] Everything is here
- [x] The code is readable
- Moving forward we probably want to refactor "downloaded data" to "cache" as a keyword in function names. New APIs use "cache".
- [x] The code is maintainable
- There are probably more possibilities for test cases, but I tried to do the most important ones, and this adds tests to previously untested features.
- test_forcingdata.py does now require downloading the data to run, but this is better than before as it tests the full workflow instead of being sensitive to current disk state
- [x] The code follows guidelines

docs/usage.rst
```diff
@@ -38,6 +38,22 @@ It is also possible to delete and re-download the data archive, in case somethin
...
+and store data into this folder. If unset, by default the package will use a
+subdirectory of the user's home folder.
+
+The download location can be retrieved using `fv3config.get_cache_dir()`, and set
+manually using `fv3config.set_cache_dir()`. If the target is set to a directory
+that already contains the archive download, it will automatically start using those
+files. Conversely, if the target is set to an empty directory, it will be necessary
+to re-download the cache files. Do not set the cache directory to a location
+that already contains non-cache files, or the files will not download until you
+call `refresh_downloaded_data` (which will delete any files in the directory).
```
oliverwm1: I don't think we should allow this behavior (deleting all files in the cache dir even if they are not related to fv3config). Why not have `FV3CONFIG_CACHE_DIR` specify the parent directory and have fv3config still control the actual name of the directory containing the input data?

fv3config/_datastore.py
```diff
@@ -57,9 +65,9 @@ def get_orographic_forcing_directory(config):
     """Return the string path of the orographic forcing directory specified by a config dictionary.
     """
     resolution = get_resolution(config)
-    dirname = os.path.join(local_archive_dir, f'orographic_data/{resolution}')
+    dirname = os.path.join(LOCAL_ARCHIVE_DIR, f'orographic_data/{resolution}')
```
oliverwm1: Should this be `get_cache_dir()`? Or why do you sometimes use `LOCAL_ARCHIVE_DIR` and sometimes `get_cache_dir()`?

fv3config/_datastore.py
```diff
@@ -2,37 +2,45 @@
...
-local_archive_filename = os.path.join(app_data_dir, filename)
-gs_bucket_prefix = 'gs://'
-
-forcing_options_dict = {
-    'default': os.path.join(local_archive_dir, 'base_forcing')
+
+if 'FV3CONFIG_CACHE_DIR' in os.environ:
+    LOCAL_ARCHIVE_DIR = os.environ['FV3CONFIG_CACHE_DIR']
+else:
+    LOCAL_ARCHIVE_DIR = os.path.join(
```
ofuhrer: If using the default, you impose some sub-structure to this directory (using appdirs). Would it make sense to impose the same substructure to the user-specified directory?

tests/test_datastore.py
```diff
@@ -0,0 +1,89 @@
+import unittest
```
ofuhrer: Very cool!

fv3config/_datastore.py
```diff
@@ -125,51 +133,50 @@ def gsutil_is_installed():
...
+            download_data_archive(archive_file)
+            archive_file.flush()
+            extract_data(archive_file.name)
 
 
 def refresh_downloaded_data():
     """Delete the cached data (if present) and re-download it."""
-    os.remove(local_archive_filename)
-    shutil.rmtree(app_data_dir)
+    shutil.rmtree(get_cache_dir())
```
ofuhrer: I always wonder how to protect something like that from cleaning away my harddrive in case get_cache_dir() returns something whacky. Can we impose some safety here? This probably holds true for all rmtree() calls in the code.

fv3config/__init__.py
```diff
@@ -7,13 +7,10 @@
...
+    refresh_downloaded_data, ensure_data_is_downloaded, set_cache_dir, get_cache_dir
+)
 
-__all__ = [
-    'config_to_namelist', 'config_from_namelist', 'get_default_config',
-    'write_run_directory',
-    'InvalidFileError', 'ConfigError'
-]
 
 __author__ = """Vulcan Technologies, LLC"""
```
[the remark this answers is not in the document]
mcgibbon: Good catch.

fv3config/__init__.py
```diff
@@ -7,13 +7,10 @@
     write_run_directory, enable_restart
 )
 from ._exceptions import InvalidFileError, ConfigError
-from ._datastore import refresh_downloaded_data, ensure_data_is_downloaded
+from ._datastore import (
+    refresh_downloaded_data, ensure_data_is_downloaded, set_cache_dir, get_cache_dir
```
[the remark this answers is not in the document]
mcgibbon: Curiosity? People asked for this, it's mostly meant for developers but our users are mainly developers in some capacity. They may want to, for example, run `du -h` in the directory to see how big the files are.

fv3config/_datastore.py
mcgibbon: I think we've tried to put as many warnings as we can on the documentation of `refresh_downloaded_data` about it deleting cached files. There are unit tests on this functionality, so the main risk is that someone sets the cache dir to something wacky, but now we're also appending `fv3config-cache` to the directory they set (or to the environment variable), so I think we're pretty safe.

On top of *all* that (remembering that this is a private function, which can only be accessed through `python3 -m fv3config.refresh-data`), I've added some interactive statements when users run the refresh command that will warn them this directory may be deleted, and given an opportunity to cancel. We don't have any current use case that requires the refresh to be non-interactive.

To some extent, the only way to be completely safe against wiping your data is backups. Barring that, I think we're now at a pretty high level of safety.

fv3config/_datastore.py
mcgibbon: I've changed it to uniformly use `get_cache_dir()`.

fv3config/_datastore.py
mcgibbon: I don't think so, the appdirs arguments are just used to find the system-located directory for specific uses (like `/username/.cache/vulcan/fv3gfs/`). Within a directory the user has specified to use for fv3config cache, it shouldn't be necessary to specify that this is fv3config or made by vulcan. We only have any kind of internal subdirectory for safety in edge cases (e.g. when the user uses `$SCRATCH` or something like that).

fv3config/__init__.py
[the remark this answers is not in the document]
oliverwm1: The use-case is needing to bind-mount the cache dir if running from outside docker with the forcing files symlinked.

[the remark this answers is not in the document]
ofuhrer: Ok, got it.

oliverwm1:
Looks good. I think the new approach for handling `refresh_download_data()` is sufficiently safe.

PR #31, opened 2019-11-19
added get_n_processes

mcgibbon:
Added a `get_n_processes` helper function. It includes a docstring that will show up in the autodoc documentation, and tests.

While doing this, I also moved the test handling over to pytest.

Ran over the checklist and I think it's all there.

### DIFF
--- a/fv3config/_config.py
+++ b/fv3config/_config.py
@@ -19,6 +19,11 @@ namelist_options_dict = {
     'default': os.path.join(package_directory, 'data/namelist/default.nml')
 }
 
+_DEFAULTS = {
+    'ntiles': 6,
+    'layout': (1, 1),
+}
+
 
 def get_default_config():
     """Returns a default model configuration dictionary."""
@@ -31,6 +36,16 @@ def get_default_config():
     config['experiment_name'] = 'default_experiment'
 
     return config
+
+
+def get_n_processes(config_dict):
+    """Return the number of processes required to run the model based on the
+    configured processor layout.
+    """
+    n_tiles = config_dict['namelist']['fv_core_nml'].get('ntiles', _DEFAULTS['ntiles'])
+    layout = config_dict['namelist']['fv_core_nml'].get('layout', _DEFAULTS['layout'])
+    processors_per_tile = layout[0] * layout[1]
+    return n_tiles * processors_per_tile
 
 
 def config_to_namelist(config, namelist_filename):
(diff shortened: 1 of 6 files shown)


fv3config/_config.py
```diff
@@ -19,6 +19,11 @@
     'default': os.path.join(package_directory, 'data/namelist/default.nml')
 }
 
+_DEFAULTS = {
```
oliverwm1: Should this be named so that it's clear it refers to namelist defaults?

fv3config/_config.py
```diff
@@ -33,6 +38,16 @@ def get_default_config():
     return config
 
 
+def get_n_processes(config_dict):
```
oliverwm1: Standard variable name is ``config`` not ``config_dict``

More generally, is there a good reason to be passing the full config dict to functions that only need a small piece of them? We are doing this throughout the code, and it's made unit testing hard for me in certain cases. We can discuss offline or in a separate issue.

fv3config/_config.py
```diff
@@ -33,6 +38,16 @@ def get_default_config():
...
+    """Return the number of processes required to run the model based on the
+    configured processor layout.
+    """
```
oliverwm1: Public methods should have complete google-style doc strings (even if they're simple functions) so that their API is clear to the user.

tests/test_n_processes.py
```diff
@@ -0,0 +1,24 @@
+import fv3config
+import pytest
```
oliverwm1: Why did you use pytest for this test, and do we want to move other tests to pytest?

fv3config/_config.py
mcgibbon: This *is* a google-style docstring. Under the style guide [Section 3.8.3](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods), I think this counts as a case "where the function's name and signature are informative enough that it can be aptly described using a one-line docstring".

tests/test_n_processes.py
mcgibbon: For the tests I needed to write, pytest let me parameterize it in a very concise way, so I went for that. pytest can run unittest style tests, so it's not necessary to change the existing tests in any way.

fv3config/_config.py
oliverwm1: Good point, I misread the style guide.

PR #34, opened 2019-11-22
create fv3run entry point

mcgibbon:
Implements `fv3run` for running the python wrapper model.

The documentation describes its usage. Here's the command line version:
```
    usage: fv3run [-h] [--runfile RUNFILE] [--dockerimage DOCKERIMAGE]
                  [--keyfile KEYFILE]
                  config outdir
```

There's also a python API with autodoc.

Tests were written for Circle CI that depend neither on docker nor on `mpirun`. There's an additional integration test that will run if docker is installed on the local machine, and the `fv3gfs-python` docker image is available. Tests will also run that use `mpirun` if it is available. These should be run locally, since it's difficult to set them up to run on Circle CI. These extra tests will automatically skip if the host machine does not have the requirements to run the test.

Other API changes:

- `config_from_namelist` now returns a pure nested dictionary with no `Namelist` objects. This results in a much cleaner yaml representation, and makes no difference to performance (only stylistic information about the namelist is lost). One side-effect is the ordering of saved namelists is now arbitrary.

Some small fixes/changes:

- `pip3` and `python3` are now used explicitly, in place of `pip` and `python`.
- pytest is now used for testing
- tests now properly revert the fv3config data storage directory at the end of each test


nbren12:
[Review FV3run PRS](https://trello.com/c/EJcMH3Mm/148-review-fv3run-prs)

nbren12:
[Review FV3run PRS](https://trello.com/c/EJcMH3Mm/148-review-fv3run-prs)

mcgibbon:
OK, checklist completed and ready for review!

docs/usage.rst
```diff
@@ -85,6 +83,49 @@ The ``namelist`` item is special in that it is explicitly stored in the ``config
...
+    optional arguments:
+      -h, --help            show this help message and exit
+      --runfile RUNFILE     location of python script to execute with mpirun
+      --dockerimage DOCKERIMAGE
+                            if passed, execute inside a docker image with the
+                            given name
+      --keyfile KEYFILE     google cloud storage key to use for cloud copy
+                            commands
+
+The only required inputs are ``config``, which specifies a yaml file containing the
```
nbren12: These docs are great. It would be nice to have a working end-to-end example and perhaps a default docker image so that people don't need to specify that manually. Is there a GCR image for fv3gfs-python?

fv3config/_config.py
```diff
@@ -60,12 +78,20 @@ def config_from_namelist(namelist_filename):
         InvalidFileError: if the specified filename does not exist
     """
     try:
-        return_dict = f90nml.read(namelist_filename)
+        return_dict = _to_nested_dict(f90nml.read(namelist_filename).items())
```
nbren12: Would [f90nml.Namelist.todict](https://f90nml.readthedocs.io/en/latest/usage.html#f90nml.namelist.Namelist.todict) work here? If so, then you could delete `_to_nested_dict`.

fv3config/_datastore.py
```diff
@@ -12,6 +12,8 @@
     USER_CACHE_DIR = os.environ['FV3CONFIG_CACHE_DIR']
 else:
     USER_CACHE_DIR = appdirs.user_cache_dir('fv3gfs', 'vulcan')
+    if not os.path.isdir(USER_CACHE_DIR):
```
nbren12: I would use `os.makedirs(USER_CACHE_DIR, exists_ok=True)` instead of this if statement.

fv3config/_datastore.py
```diff
@@ -40,6 +42,10 @@ def set_cache_dir(parent_dirname):
 
 
 def get_cache_dir():
+    return USER_CACHE_DIR
+
+
+def _get_cache_subdir():
     return os.path.join(USER_CACHE_DIR, CACHE_PREFIX)
```
nbren12: Could you describe the motivation behind this change? It is not clear to me what is a `_cache_subdir` from the name alone.

fv3config/fv3run/_run.py
```diff
@@ -0,0 +1,32 @@
+import os
+from ._docker import _run_in_docker
+from ._native import _run_native
+
+
+def run(config_dict_or_location, outdir, runfile=None, docker_image=None, keyfile=None):
```
nbren12: It would be nice if there were a default docker image, but then you would need a different way to specify that the run should occur on docker.

fv3config/fv3run/_native.py
```diff
@@ -0,0 +1,109 @@
+import logging
+import contextlib
+import resource
+import subprocess
+import os
+import tempfile
+import warnings
+import yaml
+from .._config import write_run_directory, _get_n_processes, _write_config_dict
+from ._gcloud import _copy_directory, _copy_file
```
nbren12: It seems a little odd to import a private method from a private module. Perhaps `_copy_directory` shouldn't be underscored.

tests/test_fv3run.py
```diff
@@ -0,0 +1,243 @@
...
+def count_executed_ranks(rundir):
+    filenames = os.listdir(rundir)
+    total = 0
+    while f'rank_{total}' in filenames:
+        total += 1
+    return total
+
+
+@contextlib.contextmanager
+def mocked_check_call():
```
nbren12: It seems a little complicated to write a context manager to mock out dependences. Both `unittest` and `pytest` have recommend methods for doing this. See https://docs.pytest.org/en/latest/monkeypatch.html or `unittest.patch`. Would that work instead of this context manager?

tests/test_fv3run.py
```diff
@@ -0,0 +1,243 @@
...
+)
+def test_fv3run_without_mpi(runner):
+    # the test for docker_run *will* use MPI inside the docker container
+    # we're only making it so subprocess calls of mpirun don't execute
+    if runner == docker_run and not DOCKER_ENABLED:
+        pytest.skip(f'docker or docker image {DOCKER_IMAGE_NAME} is not available')
+    fv3config.ensure_data_is_downloaded()
+    outdir = os.path.join(TEST_DIR, 'outdir')
+    with mocked_check_call(), cleaned_up_directory(outdir):
+        runner(fv3config.get_default_config(), outdir)
```
nbren12: There are no assert statements in this test. What exactly is it testing?

tests/test_fv3run.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+@pytest.mark.skipif(not MPI_ENABLED, reason='mpirun must be available')
+@pytest.mark.parametrize(
+    "runner",
+    [subprocess_run, config_dict_module_run, config_dict_filename_run]
+)
+def test_fv3run_with_mpi(runner):
```
nbren12: TBH, I don't think it is necessary to do an "end-to-end" test. I think simply ensuring that this code writes the appropriate directory and creates the correct call to `subprocess` is enough.

tests/test_fv3run.py
SpiderMonkey1975: What if we are using a platform that doesn't use an explicit call to mpirun?  The Crays will use srun to invoke MPI execution.  Do we expect to use this code on the supercomputers?

SpiderMonkey1975: Should we add MPI version checks here?  As there are ABI compatibility issues with certain versions.

docs/usage.rst
mcgibbon: Thanks! It's basically designed to work with the fv3gfs-python docker image, so perhaps that could be default... but if users have to specify an option anyways to enable docker then I figured why not do it by having them say what image they want to use? fv3gfs-python should be on GCR.

I was told that if you try to docker run an image that has a remote url it will fetch from the remote if you don't have it locally, but in practice I've found the docker run command fails whenever that happens saying you don't have the image locally. Am I doing something wrong?

nbren12: It should download. I think it could be because the image doesn't exist in GCR yet.

tests/test_fv3run.py
mcgibbon: Short answer: not much! These tests are a bit of a mess. Overhauling now.

tests/test_fv3run.py
mcgibbon: I agree. On Circle CI, only these mocked tests will run. Still, it's nice to have the end-to-end tests available locally. When they run, they do not depend on having the python wrapper installed (even in the docker container), since they use a mock runfile.

mcgibbon: In other words yeah these tests probably aren't absolutely needed and won't be part of CI where we don't have mpirun, but they helped me write the code and I think they're good to keep around.

docs/usage.rst
mcgibbon: I presume this fixed itself when you asked me to manually push an image to GCR?

fv3config/_config.py
nbren12: I think this could be done by registering a `yaml` serializer for orderdicts, but that is probably more complicated than this approach TBH.

fv3config/_datastore.py
nbren12: Maybe `_internal_cache_dir` would work. Or just copy-paste your last comment into that  the docstring of `_cache_subdir`.

mcgibbon: My hope was that "_" would make the "internal" part obvious. Nonetheless, I've renamed the function to `_get_internal_cache_dir`. It's a bit of a long name now which detracts from readability a bit, but the explicitness probably cancels that out.

docs/usage.rst
nbren12: I can know use the image, but still cannot execute an end-to-end workflow:
```
In [1]: import fv3config                                                                                                                                 

In [2]: config =fv3config.get_default_config()                                                                                                           

In [3]: fv3config.run(config, 'test_experiment', docker_image='us.gcr.io/vcm-ml/fv3gfs-python')                                                          
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/fv3gfs-python/external/fv3config/fv3config/fv3run/__main__.py", line 35, in <module>
    sys.exit(main())
  File "/fv3gfs-python/external/fv3config/fv3config/fv3run/__main__.py", line 31, in main
    run(args.config, args.outdir, args.runfile, args.dockerimage, args.keyfile)
  File "/fv3gfs-python/external/fv3config/fv3config/fv3run/_run.py", line 15, in run
    _run_native(config_dict_or_location, outdir, runfile=runfile)
  File "/fv3gfs-python/external/fv3config/fv3config/fv3run/_native.py", line 29, in _run_native
    localdir, n_processes, runfile_name=_get_basename_or_none(runfile))
  File "/usr/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/fv3gfs-python/external/fv3config/fv3config/fv3run/_native.py", line 51, in _temporary_directory
    _copy_directory(tempdir, outdir)
  File "/fv3gfs-python/external/fv3config/fv3config/fv3run/_gcloud.py", line 20, in _copy_directory
    _copy_file(source, dest, fs)
  File "/fv3gfs-python/external/fv3config/fv3config/fv3run/_gcloud.py", line 29, in _copy_file
    shutil.copy2(source_filename, dest_filename, follow_symlinks=True)
  File "/usr/lib/python3.7/shutil.py", line 266, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.7/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
PermissionError: [Errno 13] Permission denied: '/outdir/co2historicaldata_2012.txt'
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
<ipython-input-3-3bb6607f5ad4> in <module>
----> 1 fv3config.run(config, 'test_experiment', docker_image='us.gcr.io/vcm-ml/fv3gfs-python')

~/workspace/fv3config/fv3config/fv3run/_run.py in run(config_dict_or_location, outdir, runfile, docker_image, keyfile)
     27         _run_in_docker(
     28             config_dict_or_location, outdir, docker_image,
---> 29             runfile=runfile, keyfile=keyfile
     30         )
     31     else:

~/workspace/fv3config/fv3config/fv3run/_docker.py in _run_in_docker(config_dict_or_location, outdir, dockerimage, runfile, keyfile)
     28         subprocess.check_call(
     29             DOCKER_COMMAND + bind_mount_args + docker_args + [dockerimage] +
---> 30             python_command + python_args)
     31 
     32 

~/miniconda3/envs/fv3net/lib/python3.7/subprocess.py in check_call(*popenargs, **kwargs)
    345         if cmd is None:
    346             cmd = popenargs[0]
--> 347         raise CalledProcessError(retcode, cmd)
    348     return 0
    349 

CalledProcessError: Command '['docker', 'run', '-v', '/tmp/tmpepr2szo0.yaml:/fv3config.yaml', '-v', '/home/noahb/workspace/fv3config/examples/test_experiment:/outdir', '--user', '1000:1000', 'us.gcr.io/vcm-ml/fv3gfs-python', 'python3', '-m', 'fv3config.fv3run', '/fv3config.yaml', '/outdir']' returned non-zero exit status 1.
```

fv3config/_config.py
nbren12: Ok. I guess it's okay. Although `yaml` does have a way to register a specific processor for data types, I think this approach is cleaner.

fv3config/_datastore.py
nbren12: ok. thanks.

fv3config/fv3run/_docker.py
```diff
@@ -0,0 +1,68 @@
...
+            config_location = DOCKER_CONFIG_LOCATION
+    else:
+        _write_config_dict(config_dict_or_location, config_tempfile.name)
+        bind_mount_args += ['-v', f"{config_tempfile.name}:{DOCKER_CONFIG_LOCATION}"]
+        config_location = DOCKER_CONFIG_LOCATION
+    return config_location
+
+
+def _get_docker_args(docker_args, bind_mount_args, outdir):
+    bind_mount_args += ['-v', f"{os.path.abspath(outdir)}:{DOCKER_OUTDIR}"]
```
nbren12: I would add `--rm` so that he containers are deleted once they run.

fv3config/fv3run/_run.py
nbren12: Hmm. I wonder if this top-level `run` function is a little too inflexible for all use-cases. For instance, there is currently no way to add flags to the docker command. I would suggest making `_run_native` and `_run_in_docker` public methods since those will be easier to extend.

fv3config/fv3run/_native.py
nbren12: Okay. I always thought the `_` meant private to a module, rather than package-private, but it's all just a convention with no teeth anyway.

tests/test_fv3run.py
nbren12: no problem!

tests/test_fv3run.py
nbren12: Hmm. Ok. In general, I think we need a better separation between unittests which should ideally only interact with mock objects, and integration or regression tests which can interact with the outside world. Having too many tests calling subprocesses, interacting with the filesytem, or making HTTP requests can slow things down and make them less portabale. I also think it can indicate some issues with the design. That isn't the case for this PR, but has been for others.

tests/test_fv3run.py
```diff
@@ -0,0 +1,252 @@
...
+        assert mock.call_args[0] == ([
+            'mpirun', '-n', '6', 'python3', '-m', 'mpi4py', 'mock_runscript.py'],)
+        config = yaml.safe_load(open(os.path.join(outdir, 'fv3config.yml'), 'r'))
+        assert config == fv3config.get_default_config()
+
+
+def test_fv3run_docker():
+    """End-to-end test of running a mock runscript inside a docker container.
+    """
+    if not DOCKER_ENABLED:
```
nbren12: Not necessary to change, but `pytest` has a [skipif decorator][1] that does this.

[1]:  http://doc.pytest.org/en/latest/skipping.html#id1

mcgibbon: Updated. This was left over from when the tests were bundled in a way that didn't allow using the decorator.

fv3config/fv3run/_docker.py
mcgibbon: Added.

fv3config/fv3run/_run.py
mcgibbon: This run function is meant to replicate and be the only call for the command-line API, and I think most methods of running in docker or native are ones we want to expose in the command-line API. In other words, I think making `run_native` and `run_in_docker` public doesn't remove the need to extend `run` to support this behavior. And the code of `run` itself is easy to extend because it separates concerns by passing along to `_run_native` or `_run_in_docker` immediately.

Part of the zen of Python is that there should be ideally one obvious way of doing a thing. It's unclear to me what the user gains from having either e.g. both `run` or `run_native` to run the model natively. If it seems cleaner to have `run_docker` and `run_native` *instead* of `run`, and make `run` a private interface called by the command-line endpoint, that might be a nice route to go.

mcgibbon: Especially given I'm already finding I don't really want to add kubernetes into the `run` endpoint, because it adds *lots* of new options.

PR #35, opened 2019-11-27
replaced gsutil usage with gcsfs

mcgibbon:


fv3config/_asset_list.py
```diff
@@ -102,71 +116,45 @@ def get_asset_dict(source_location, source_name, target_location='',
...
         target_location (str, optional): target_location used for generated assets.
             Defaults to '' which is root of run-directory.
+        copy_method ('copy' or 'link', optional): whether to copy or link assets,
+            defaults to 'copy'. If location is a google cloud storage url, this option
+            is ignored and files are copied.
 
     Returns:
         list: a list of asset dictionaries
         """
+    if is_gsbucket_url(location):
```
nbren12: I would recommend creating the filesystem object using the approach from my multiple restart PR: https://github.com/VulcanClimateModeling/fv3net/blob/7c224ffc885f15196c64017ea941784e53af5d93/external/vcm/vcm/misc/rundir.py#L81 

This will allow using any type of filesystem supported by `fsspec` rather than just GCS or local. `fsspec.filesystems('file')` creates a local file system, `fsspec.filesystems('gs')` creates a GCS one.

fv3config/_asset_list.py
```diff
@@ -102,71 +116,45 @@ def get_asset_dict(source_location, source_name, target_location='',
...
-    path_list = stdout_str.decode().split('\n')[:-1]
-    for path in path_list:
-        dirname, basename = os.path.split(path)
-        if basename != '':
-            asset_list.append(get_asset_dict(dirname,
-                                             basename,
-                                             target_location=target_location))
+    path_list = fs.walk(location)
+    for dirname, _, files in path_list:
+        print(dirname, files, target_location)
```
nbren12: perhaps removing this print statement, which i assume was there for debugging purposes.

fv3config/_asset_list.py
```diff
@@ -221,28 +214,25 @@ def config_to_asset_list(config):
...
+            'cannot perform linking operation involving remote urls '
+            f'from {source_item} to {target_item}'
+        )
     if os.path.exists(target_item):
         os.remove(target_item)
     os.link(source_item, target_item)
 
 
 def copy_file(source_path, target_path):
     if is_gsbucket_url(source_path):
```
nbren12: Same trick here as I describe above to create the filesystem object in a general fashion.

fv3config/_asset_list.py
```diff
@@ -102,71 +116,45 @@ def get_asset_dict(source_location, source_name, target_location='',
...
-                                             basename,
-                                             target_location=target_location))
+    path_list = fs.walk(location)
+    for dirname, _, files in path_list:
+        print(dirname, files, target_location)
+        subdir_target_location = os.path.join(
+            target_location, os.path.relpath(dirname, start=location)
+        )
+        for basename in files:
+            asset_list.append(
```
nbren12: In `gcsfs`, there is no `gs://` prefix. Are you putting them back in anywhere? If not, then I do not see how the generated `asset` will not contain any information about whether the file is remote or local. 

However, instead of re-prepending the `gs://` to the path, I would recommend adding a `protocol` entry to the asset, which can be used to instantiate the filesystem in `copy_file`. To write a file this would look like:
```
def copy_file(protocol, source_path, target_path):
    fs = fsspec.filesystems(protocol)
    fs.get(source_path, target_path)
```
This would change the asset API, but so far I don't think any of us are using that directly.

fv3config/_asset_list.py
nbren12: `fsspec.filesystems` is really nice example of using the factory design pattern IMO.

fv3config/_asset_list.py
mcgibbon: Yes, I ran into that bug when working on the fv3run kubernetes PR and prepended it there. Part of why I didn't ask for reviews on this PR yet was that the gcs functionality couldn't be tested until the kubernetes stuff was working, and now we have that as an integration test (and a hint as to at least one useful unit test I can add).

The main reason I'd opt for pre-pending gs:// instead of adding `protocol` is it makes the two APIs (location and asset) consistent with one another. Because the asset API *is* a user-facing api (even if none of us are using it currently), I think consistency of the APIs is worthwhile. It's also not hard to enforce or make simple to access/generalize, we could write a `get_protocol(path)` function (I effectively did this, but `get_fs(path)`, in the upcoming kubernetes API).

Looking at the changes in the kubernetes branch, the code has actually been changed a fair bit more -- I think it may be best to review them as a single PR instead of working to pull just the filesystem changes out.

mcgibbon: It also ends up being a little more complicated than picking one protocol, because there are separate `put`, `get`, and `copy` methods for copying to, from, or within a filesystem. The code I wrote in the next PR handles these different cases.

PR #37, opened 2019-12-03
Create run_kubernetes function

### LINKED ISSUE #33: Use gcsfs instead of gsutil for interacting with google cloud storage
We should move to using the `gcsfs` package for interacting with google cloud storage. This will make generating an asset list from a google cloud storage url more straightforward (see [here](https://github.com/VulcanClimateModeling/fv3config/blob/b84c948b4563408915b88df7efe6bbfb02b64c4e/fv3config/_asset_list.py#L145-L167) for current implementation with `gsutil`). With `fsspec` (https://filesystem-spec.readthedocs.io/en/latest/intro.html) it should be possible to use the same code for the local and remote filesystems.

mcgibbon:
We need to be able to run the model on kubernetes. This PR creates a python function `run_kubernetes` for doing so. It has not yet been added to the bash `fv3run` command.

Fixes #33.

Other API changes:

- `DependencyError` is removed, instead a `DelayedImportError` object is used to delay the `ImportError` until the code attempts to access a member of the imported module.

Non-API/internal changes:

- pytest is now used to run tests
- `make coverage` now uses pytest
- python 3 versions of python binaries are called explicitly (e.g. `python3` and `pip3`)
- internally the package now uses gcsfs instead of gsutil for copy operations. `fsspec`, which has the same API as `gscfs`, is now used for local copy operations other than linking
- added an example for creating a run directory using an external path, which I used to manually test remote copy operations
- asset list generation for remote and local directories has been unified into one function
- tree walking responsibility and asset creation responsibility of asset list generation has been separated into two functions
- distinction is made between "get" and "copy" in file operation helper function names
- native fv3run functionality uses a named logger instead of the root logger
- removed unused copy/link methods in `_datastore.py`
- fixed some code-formatting in sphinx docs


docs/usage.rst
```diff
@@ -142,6 +142,38 @@ separate functions based on where the model is being run.
...
+A python interface is provided for submitting `fv3run` jobs to Kubernetes. Here's
+an example for submitting a job based on the default configuration dictionary::
+
+    import yaml
+    import gcsfs
+    import fv3config
+
+    config_location = 'gs://my_bucket/fv3config.yml'
+    outdir = 'gs://my_bucket/rundir'
+    docker_image = 'us.gcr.io/vcm-ml/fv3config-python'
```
nbren12: This image has not been pushed to GCS, so I couldn't test whether this worked or not.

mcgibbon:
Ah sorry `fv3config-python` is a typo, it should be `fv3gfs-python`.

docs/usage.rst
```diff
@@ -142,6 +142,38 @@ separate functions based on where the model is being run.
...
+
+    fs = gcsfs.GCSFileSystem()  # project name is optional,
+                                # we don't use commands that depend on it
+    with fs.open(config_location, 'w') as config_file:
+        config_file.write(yaml.dump(config))
```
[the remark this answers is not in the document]
mcgibbon: Done.

docs/usage.rst
```diff
@@ -142,6 +142,38 @@ separate functions based on where the model is being run.
...
+
+    fv3config.run(
```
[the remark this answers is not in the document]
mcgibbon: Done.

docs/usage.rst
```diff
@@ -142,6 +142,38 @@ separate functions based on where the model is being run.
 .. autofunction:: fv3config.run_native
 .. autofunction:: fv3config.run_docker
 
+Submitting a Kubernetes job
+---------------------------
+
+A python interface is provided for submitting `fv3run` jobs to Kubernetes. Here's
```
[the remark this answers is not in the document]
mcgibbon: I don't think that would be appropriate to put here. Users can use kubectl or possibly the web console to get this info. I will however have the example use `gcp-key` as the secret name.

fv3config/filesystem.py
```diff
@@ -0,0 +1,59 @@
...
+def _is_gcloud_path(path):
+    return path.startswith(GS_BUCKET_PREFIX)
+
+
+def _get_gcloud_project():
+    return os.environ.get(google.auth.environment_vars.PROJECT)
+
+
+def _get_fs(path):
+    if _is_gcloud_path(path):
```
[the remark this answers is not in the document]
mcgibbon: I think that would be appropriate when we decide we want to support more than just gcs as a remote file system. As it stands, the current code is easier to read than this example and does the same thing in half the number of lines. But certainly it would be cleaner to call fsspec.filesystem(proto) when proto can take on more than one possible value (currently only 'gs').

Supporting alternate remote filesystems will be significantly more complicated than just accessing additional fsspec options - authentication has been by far most of the work in getting gcsfs to work.

docs/usage.rst
mcgibbon: Typo has been fixed, it's `fv3gfs-python`.

fv3config/filesystem.py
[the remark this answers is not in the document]
mcgibbon: I would say though in the long run that's probably the way we want to go, and we can ensure it will be easy to switch over to that usage by always instantiating the fs object using `_get_fs()`.

docs/usage.rst
```diff
@@ -142,6 +142,40 @@ separate functions based on where the model is being run.
...
+        docker_image,
+        gcp_secret='gcp-key'  # replace with your kubernetes secret
+                              # containing gcp key in key.json
+    )
+
+The gcp key is generally necessary to gain permissions to read and write from google
+cloud storage buckets. In the unlikely case that you are writing to a public bucket,
+it can be ommitted.
+
+.. autofunction:: fv3config.run_kubernetes
```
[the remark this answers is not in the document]
mcgibbon: Fixed. I also went through the whole file and made sure API links exist for all mentioned methods.

fv3config/_asset_list.py
```diff
@@ -221,28 +206,23 @@ def config_to_asset_list(config):
...
-
-def copy_file(source_path, target_path):
-    if is_gsbucket_url(source_path):
-        if gsutil_is_installed():
-            check_call(['gsutil', 'cp', source_path, target_path])
-        else:
-            raise DependencyError(f'Optional dependency gsutil not found. File {source_path} will not be copied to {target_path}')
-    else:
-        shutil.copy(source_path, target_path)
+    os.symlink(source_item, target_item)
```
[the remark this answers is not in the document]
mcgibbon: hardlinking can't work when root owns the cache directory, which is the case in our docker container. Also, the use case that required hard links which involved linking the input data in from outside the docker container no longer exists - data is either retrieved from the docker cache, or downloaded from the cloud.

tests/test_asset_list.py
```diff
@@ -179,10 +181,15 @@ def test_asset_list_from_local_dir_one_file_in_root(self):
...
         workdir = self.make_work_directory('workdir_with_files')
         self.make_empty_files(os.path.join(workdir, TEST_SOURCE_DIR), FILELIST_WITH_SUBDIR)
-        asset_list = asset_list_from_local_dir(os.path.join(workdir, TEST_SOURCE_DIR))
+        asset_list = asset_list_from_path(os.path.join(workdir, TEST_SOURCE_DIR))
         # necessary to have full path for source location in test assets
         for asset in ASSET_LIST_WITH_SUBDIR:
             asset['source_location'] = os.path.join(workdir, asset['source_location'])
+        for item in asset_list:
+            print(item)
+        print('reference')
```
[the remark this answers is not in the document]
mcgibbon: Nope. Removed.

fv3config/filesystem.py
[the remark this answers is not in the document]
nbren12: I don't think it's harder to read, and it would allow you to remove a significant chunk of the code in this module. It would allow you to delete `_is_gcloud_path`, `_get_gcloud_fs`, `_get_gcloud_project`, as well as the manual import and instantiation `LocalFileSystem`. I see no reason not to use the more general purpose solution in this case.

docs/usage.rst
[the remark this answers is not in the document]
nbren12: Why not? Currently we are the only users of this package. I don't see why we can't add some information that will make our lives easier.

tests/test_kubernetes.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def test_local_outdir_rejected(config_location, docker_image, runfile):
+    with pytest.raises(ValueError):
+        fv3config.fv3run._kubernetes._get_job(
+            config_location, '/local/outdir', docker_image, runfile)
+
+
+def test_get_job(
+        config_location, outdir, docker_image, runfile, jobname,
```
nbren12: This test is bit tricky to read, some whitespace and comments would help, but it would be better to split it into several functions. There are a lot of assert statements for one tests.

tests/test_kubernetes.py
```diff
@@ -0,0 +1,166 @@
...
+        _check_env(container.env, 'CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE', '/secret/gcp-credentials/key.json')
+    else:
+        _check_env(container.env, 'GOOGLE_APPLICATION_CREDENTIALS', None)
+        _check_env(container.env, 'CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE', None)
+    _check_env(container.env, 'GOOGLE_CLOUD_PROJECT', google_cloud_project)
+    _check_resource_requirements(
+        container, memory_args.memory_gb, memory_args.memory_gb_limit, cpu_count)
+
+
+def _check_job(job, jobname):
```
nbren12: I think you should change "check" to "assert" in all these functions. In this case, "check" and "assert" have the same meaning, so it is confusing that we use `assert` in some places, and `check` in others. *Clean code* makes the point that the same concept shouldn't have two names.

tests/test_asset_list.py
```diff
@@ -104,6 +104,8 @@ def cleanup(self):
 
 class AssetListTests(unittest.TestCase):
 
+    maxDiff = 999
```
nbren12: What does this mean?

setup.py
```diff
@@ -45,8 +46,9 @@
     },
     install_requires=requirements,
     extras_require={
-        'bucket-access': 'gsutil',
+        'bucket-access': ['gcsfs', 'google-cloud-storage'],
```
nbren12: is `google-cloud-storage` not a dependency of `gcsfs`?

fv3config/fv3run/_kubernetes.py
```diff
@@ -0,0 +1,266 @@
...
+                name='gcp-key-secret',
+                secret=kube.client.V1SecretVolumeSource(
+                    secret_name=self.gcp_secret
+                )
+            )
+        else:
+            return None
+    
+    @property
+    def volume_list(self):
```
nbren12: Nit: maybe rename to `volumes`. I am not a huge fan of naming things with the type in them unnecessarily. Is it really important that it's a "list" versus another type of iterable?

fv3config/fv3run/_kubernetes.py
```diff
@@ -0,0 +1,266 @@
...
+        if self.gcp_secret is not None:
+            return [self._secret_volume]
+        else:
+            return []
+    
+    @property
+    def volume_mounts_list(self):
```
nbren12: Same comment as above. `volume_mounts` would be fine.

fv3config/fv3run/_kubernetes.py
```diff
@@ -0,0 +1,266 @@
...
+                    name=self._secret_volume.name,
+                    read_only=True,
+                )
+            ]
+        else:
+            volume_mounts_list = []
+        return volume_mounts_list
+
+    @property
+    def _secret_env_list(self):
```
nbren12: `env` isn't 100% clear. I would say `_secret_environmental_variables`

fv3config/filesystem.py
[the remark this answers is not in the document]
nbren12: I think this module is really a good demonstration of why there are design patterns (e.g. the factory pattern, prototype pattern, etc) for creating objects.

docs/usage.rst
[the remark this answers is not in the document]
mcgibbon: I added the note to the `fv3net` readme that our vcm key name is `gcp_key`. I think that is the "right" place to put internal onboarding documentation as to how our workflows work and how our credentials are set up. That coupled with the example value being `gcp_key` should be enough to tell people what key to use.

On top of that, `kubectl get secrets` immediately says we have a gcp_key secret.

Even if we're the only people using this package right now, the idea is to make it so that's not the case. We have some steps to get there still but I'd like to step forward. Personally I'd be afraid to use a package that goes as far as to say what their internal credentials secret is called in the package documentation.

fv3config/filesystem.py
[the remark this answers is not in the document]
mcgibbon: I've combined what you're saying about being able to delete _is_gcloud_path etc. with what I was saying about writing the code for only `gs://` at the moment. Everything is now completely agnostic to the type of remote filesystem being used, so adding remote filesystems will really be as simple as modifying `get_fs`.

I also felt `get_fs` is a central enough method that I removed the underscore, even if it's meant to be used internally only and probably shouldn't be imported into the top-level namespace. Its API should remain stable.

tests/test_kubernetes.py
mcgibbon: They are distinct concepts. You "check" a noun while you "assert" a statement. "assert job" implies asserting bool(job), in other words that job exists. I think I have been consistent in using "assert_{expression}" everywhere else, "check_{object}" has generally not been necessary because we've avoided writing many objects.

I really dislike having a "check_{object}" test at all because it implies we're testing multiple expressions about the object and I'd rather have one test for each, but there are way too many super simple tests for that to be good here.

It's also generally bad to "test that the object you instantiated is what you supplied as arguments", because it generally implies the object you instantiated should be simpler. It's essential here because we're wrapping some pretty gnarly google API we aren't allowed to change into a much simpler job creation API.

I could have used "_assert_job_is_correct", but I felt "_check_job" was a more succinct way of saying this. Would you prefer I change the structure of the names to `_assert_{noun}_is_correct`?

tests/test_asset_list.py
mcgibbon: It's a pytest thing that will increase the maximum length of a diff that it will show. I added it to get the output I needed to see at the moment a test was failing and it *could* be removed, but it probably should be kept in because apparently the asset list tests are such that they make errors with very large diffs that need to be printed.

tests/test_kubernetes.py
mcgibbon: Hahaha you should have seen it before I refactored the code into 6 different functions.

The core issue is that `get_job` has a lot of arguments. In order to test all possible combinations, this test takes 9 fixtures. The main test function actually has fewer than 2 lines per fixture, and I think has gotten to be about as efficient/readable as it can be while still fully testing the 9-argument function.

I squeezed out a little more readability by combining the secret-related checks into a helper.

The "baseline" for testing this was previously there were zero tests and people didn't seem to think having this test was necessary. Now there is a test that will find pretty much all the errors, though it's less than optimally readable (to be generous). It can probably be improved by splitting it somehow into multiple tests, but I think that's more effort than required here.

setup.py
mcgibbon: I had added it as an explicit dependency because I was using it explicitly to get the environment variable name used for the gcloud project name. No longer doing that, so I removed the dependency.

fv3config/fv3run/_kubernetes.py
mcgibbon: I used `env` because that is the name used by the kubernetes API.

fv3config/fv3run/_kubernetes.py
mcgibbon: Having both `varname` and `varnames` is a specific example in _Clean Code_ of something not to do, since it's hard to differentiate the two at a glance, and easy to accidentally typo one into the other. As a result I often append `_list` to an iterable of things, that does not necessarily need to be of type `list`.

Here we don't have both names, and the name used by kubernetes is also `volumes` and `volume_mounts`, so I've renamed it as you suggested.

fv3config/fv3run/_kubernetes.py
mcgibbon: Done.

tests/test_kubernetes.py
nbren12: Okay. That is fine. I agree that it's hard to cleanly test a deeply nested structure like the google api. So this is more of an integration test anyway.

setup.py
nbren12: Okay. thank you.

fv3config/fv3run/_kubernetes.py
nbren12: Yah maybe it's having two names which is the problem. I can't think of a case when a good software unit would have both `volume` and `volumes` attributes since that would indeed be confusing.

I don't remember that section in Clean Code. Is there something wrong with this?
```
for name in names:
   pass
```
That seems like good idiomatic python code to me.

mcgibbon: Apparently it wasn't in Clean Code, I must have read it somewhere else. I do find though that it's easier to read at a glance:
```
for name in name_list:
    pass
```
I also have been burned at least twice by using `for name in names` where I accidentally wrote `names` (or the equivalent for the name I had) within the for loop.

I agree that when they don't appear together, the simple plural noun is probably best.

PR #41, opened 2019-12-19
Add linting to makefile and Circle CI

mcgibbon:
Replicates the workflow for linting used in `fv3net`, where linting runs flake8 and also checks that black would not change any files.


fv3config/_exceptions.py
```diff
@@ -1,10 +1,12 @@
 class DataMissingError(FileNotFoundError):
     """Raised when expected cached data is not present."""
+
     pass
 
 
 class InvalidFileError(FileNotFoundError):
     """Raised when a specified file is invalid, either non-existent or not as expected."""
+
     pass
```
mcgibbon: Feels very strange that black adds a line here.

mcgibbon:
Rebased in a new branch to clean up changes from another PR.

PR #43, opened 2020-01-03
Add node pool taint toleration to kubernetes submission

frodre:
This PR adds a toleration to the kubernetes job submission process.  The toleration is keyed to allow the jobs to run on the large nodes that exist on the `ml-cluster-dev` kubernetes cluster.  The need for a taint/toleration procedure is to keep specific `kube-system` pods off of the large nodes for proper node spin-down after completion.  Prior to this, some of the pods would stay up indefinitely, which would be needlessly expensive.  

I considered formalizing this as a configuration argument in `KubernetesConfig` for letting the user specify multiple tolerations, but since we only have one simulation cluster currently and since that cluster (with the taint) will change relatively infrequently, I don't think it's necessary yet.


frodre:
[Update terraform documentation and automate VM kubectl configuration for proxy cluster access.](https://trello.com/c/cY4fHktf/181-update-terraform-documentation-and-automate-vm-kubectl-configuration-for-proxy-cluster-access)

tests/test_kubernetes.py
```diff
@@ -114,6 +114,10 @@ def test_get_job(
     _check_resource_requirements(
         container, memory_args.memory_gb, memory_args.memory_gb_limit, cpu_count
     )
+    toleration = pod_spec.tolerations[0]
+    assert toleration.effect == "NoSchedule"
+    assert toleration.key == "dedicated"
+    assert toleration.value == "climate-sim-pool"
```
mcgibbon: This would be suitable to put in a `_check_tolerations` helper function, and should additionally test that `len(pod_spec.tolerations) == 1`. After that, looks ready to merge.

frodre: Thanks! Changes have been made to address these two suggestions.

PR #45, opened 2020-01-10
Add local caching to get_file operation

mcgibbon:
This PR modifies `get_file` so that it caches remote files in the fv3config cache directory. Presently this is done with all files, and there isn't an API to disable it. Adding such an API would be fairly easy, but we'd need to come up with the design (easiest would be some kind of `cache_downloads(True/False)` routine).

Other small changes:
- functions in `derive` that pertained to filenames for options have been moved to `_datastore.py` alongside the routines that retrieve directories for options. This was necessary to avoid circular imports.
- functions for getting/setting the cache directory have been moved to their own module. This was necessary to avoid circular imports.
- anchor versions for black and flake8 in the circleci config and requirements_dev.txt
- Ignore a flake8 style convention that black and flake8 disagree on (whitespace before ":")

Tests have been added to check the location chosen to cache a file. The caching itself is untested because it requires remote url access, and might not be necessary because the function itself is quite simple (three lines).


tests/test_datastore.py
```diff
@@ -31,35 +33,35 @@ def mock_requests_get(url):
...
     def test_get_archive_dir(self):
-        result = fv3config.get_cache_dir()
+        result = fv3config.cache_location.get_cache_dir()
         assert isinstance(result, str)
         assert os.path.isabs(result)
 
     def test_set_then_get_archive_dir(self):
         with tempfile.TemporaryDirectory() as tempdir:
-            original = fv3config.get_cache_dir()
+            original = fv3config.cache_location.get_cache_dir()
```
mcgibbon: FYI changes like this were done automatically by PyCharm's refactor operation. I'm leaving it because it's fine to be explicit about where the function lives in the test files, but it does still exist at the top level and that's where it should be used in user code.

fv3config/_asset_list.py
```diff
@@ -172,7 +172,7 @@ def write_asset(asset, target_directory):
     if not os.path.exists(os.path.dirname(target_path)):
         os.makedirs(os.path.dirname(target_path))
     if copy_method == "copy":
-        get_file(source_path, target_path)
+        filesystem.get_file(source_path, target_path, cache=True)
```
mcgibbon: This is the only place I've added caching, which I think is sufficient. Let me know if that's not the case.

Issue #46, opened 2020-01-11
Remove "options" for file locations in favor of automatic cache


mcgibbon:
We should move towards a system that no longer has "options" available for data, and instead uses remote URLs that get cached locally. Motivations for this change are listed below.

This involves a few changes:

- Replace `get_default_config()` with `get_config(location)` which accepts a remote file location. Initially, get_default_config should remain in place but should present a DeprecationWarning. It would return a dictionary that uses remote paths instead of "options".
- Remove `initial_conditions` and `forcing` keys from the configuration dictionary.
- Rename `patch_files` to `file_sources`.
- Allow not just asset dictionaries but also strings in `file_sources`. A string should represent a path (either file or directory) which is copied into the run directory recursively. These would be replaced by asset dictionaries internally. To place files in subdirectories, users should either place the files under a subdirectory in the source location, or use the more extensive asset dict representation (@oliverwm1 has found for loops generating asset dicts to be a very smooth workflow).
- Add keys `orographic_data` and `field_tables` to the config dict. These locations should be directories which have the same structure as the current cache locations. Orographic data should be placed in resolution subfolders and field_tables should be labelled by scheme.
- Remove the `data_table` key, treat it as a forcing file.
- Treat "default" as a filename instead of option for `diag_table`.

A sample configuration dictionary (excluding the namelist) might look like the following yaml (note this is a mockup and doesn't represent a valid run directory):
```
experiment_name: default_experiment
forcing: gs://vcm-fv3config/data/base_forcing/v1.1
initial_conditions: gfs_example
file_sources: [
    gs://vcm-fv3config/data/base_forcing/v1.1,
    gs://vcm-fv3config/config/data_table/v1.0/data_table,
    {
        source_location: gs://vcm-fv3config/data/initial_conditions/gfs_initial_conditions/v1.0,
        source_name: file.nc,
        target_location: ,
        target_name: file.nc,
        copy_method: copy,
    },
]
orographic_data: gs://vcm-fv3config/data/orographic_data/v1.0
field_tables: gs://vcm-fv3config/config/field_tables/v1.0
diag_table: gs://vcm-fv3config/config/diag_table/v1.0/diag_table
```

Motivations:

[...]

PR #49, opened 2020-01-21
added support for bind-mounting local files in run_docker

### LINKED ISSUE #48: Let `run_docker` work with assets present on host system
It would also be nice to make `run_docker` work with assets that are present on this host system, which can be symlinked in using bind_mounts.

mcgibbon:
Fixes #48. Bind-mounts at the same level as specified by the config dict (directories if absolute path, filename if asset dict).

A side-effect of this is that calling run_docker will trigger an access request on the configuration dictionary. If the user does not have access, the command will fail (not really an issue, since it would also fail inside the docker container in that case). This did require mocking out the access in some other unit tests.

No documentation change needed as the lack of support for bind-mounting local files was undocumented. I did shift one documentation section at the suggestion of @oliverwm1 to make the order of information smoother.

Tests that the correct locations are bind-mounted are added.

### DIFF
--- a/tests/test_fv3run.py
+++ b/tests/test_fv3run.py
@@ -369,7 +369,7 @@ def test_get_credentials_args(keyfile, expected_docker_args, expected_bind_mount
                     }
                 ],
                 "data_table": "gs://bucket/data_table/default",
-                "patch_files": []
+                "patch_files": [],
             },
             [],
         ],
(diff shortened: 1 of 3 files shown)


tests/test_fv3run.py
```diff
@@ -265,3 +279,83 @@ def test_get_credentials_args(keyfile, expected_docker_args, expected_bind_mount
...
+                "/local/forcing/default",
+                "/local/diag/default",
+            ],
+        ],
+        [
+            {
+                "experiment_name": "default_experiment",
+                "initial_conditions": "gs://bucket/initial_conditions/default",
+                "forcing": "gs://bucket/forcing/default",
+                "diag_table": [
```
oliverwm1: This test is a bit confusing, since in practice `diag_table` cannot be specified as an asset dict. Perhaps this was meant to be the `patch_files` entry? Either way, a config dict that uses the `patch_files` option should be tested.

oliverwm1: The docs state that initial_conditions, forcing or patch_files can use the asset representation.

We could update the API, but I suggest we also address this as part of #46.

PR #53, opened 2020-01-22
Only use --oversubscribe flag if cpu count requires it

oliverwm1:
This PR ensures the `--oversubscribe` MPI flag is only used if the CPU count requires it. It adds the `--use-hwthread-cpus` flag so that hyperthreaded cores (such as on google VMs and MacBooks) will be used.

It also fixes a bug in `_create_job_object` by ensuring that the container name is alphanumeric.

### DIFF
--- a/fv3config/fv3run/_native.py
+++ b/fv3config/fv3run/_native.py
@@ -2,6 +2,7 @@ import logging
 import contextlib
 import resource
 import subprocess
+import multiprocessing
 import os
 import tempfile
 import warnings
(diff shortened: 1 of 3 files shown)


fv3config/fv3run/_native.py
```diff
@@ -65,6 +66,16 @@ def _set_stacksize_unlimited():
         )
 
 
+def _add_oversubscribe_if_necessary(mpi_flags, n_processes):
+    try:
+        cpu_count = multiprocessing.cpu_count()
+        if cpu_count < n_processes:
+            mpi_flags += ["--oversubscribe"]
+    except NotImplementedError:
+        warnings.warn("could not determine cpu count, assuming it is sufficient")
```
mcgibbon: replace "it is sufficient" with "number of processors is at least as many as number of MPI tasks"

oliverwm1: Done, thanks.

PR #54, opened 2020-01-28
Instructions on incrementing and tagging a version

mcgibbon:
Added a RELEASE.rst file with release instructions. For us right now that means incrementing the version, tagging a commit, and pushing it to master. Further down the line this file can include instructions on deploying to pypi.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,10 @@
+Release Instructions
+====================
+
+1. Prepare master branch for release (make sure all PRs are merged and tests pass).
+
+2. Run `bumpversion <major/minor/bugfix>`. This will create a new tagged commit,
+   having updated all version references to the new, higher version.
+
+3. Run `git push && git push origin --tags` to push the version reference commit and
+   then all local tags to master.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,7 +2,14 @@
 History
 =======
 
+
+0.2.0 (2020-01-27)
+------------------
+
+* Began tagging version commits
+
+
 0.1.0 (2019-10-11)
 ------------------
 
-* First release on PyPI.
+* Initial pre-alpha release


RELEASE.rst
```diff
@@ -0,0 +1,10 @@
+Release Instructions
+====================
+
+1. Prepare master branch for release (make sure all PRs are merged and tests pass).
+
+2. Run `bumpversion <major/minor/bugfix>`. This will create a new tagged commit,
```
oliverwm1: Is `bumpversion` a standard tool? We should probably have some info about where to install it from.

RELEASE.rst
```diff
@@ -0,0 +1,10 @@
+Release Instructions
+====================
+
+1. Prepare master branch for release (make sure all PRs are merged and tests pass).
```
oliverwm1: Does every merge to master mean an increase in minor version number? How do we decide when to increase version number? (This is probably a better discussion for offline, but maybe we could add some guidelines to our code review document?)

oliverwm1:
My main comment is that it would be helpful to have a link or some small explanation of where to get `bumpversion` from. I'm not familiar with this tool, don't have it installed on my VM or Mac. Also, have a more open-ended comment about versioning, can be discussed more offline.

RELEASE.rst
mcgibbon: It's on pypi, and it's in the requirements_dev.txt. Package name is bump2version, it behaves the same as the now unsupported bumpversion package.

RELEASE.rst
mcgibbon: No, you have to manually check out the master branch and perform this release process to increase the version number. I think it's a separate thing from code reviews, but we could have more discussion about our overall code design process. If it's plugged into our current process, we'd discuss this in the morning or during our weekly meetings.

mcgibbon:
Tests now pass on fv3gfs-python, so this is ready to merge when you approve it.

RELEASE.rst
oliverwm1: Okay sounds good, didn't realize it was in requirements_dev.txt

PR #55, opened 2020-01-28
Fix test_fv3run for oversubscribed testing system

oliverwm1:
This fixes `test_fv3run_with_mocked_subprocess` to account for the case where the testing system is oversubscribed, and hence `--oversubscribe` is created as a flag to be passed to mpirun.


mcgibbon:
Changes from this PR have been completed in #54 (for some reason the logic used here to remove the oversubscribe still didn't work).

PR #61, opened 2020-02-06
Add ability to use a custom field_table file

spencerkclark:
This adds the ability to add a custom field_table in the configuration dictionary to override the inferred version based on the microphysics scheme used.  The behavior is implemented as described in [this design document](https://paper.dropbox.com/doc/fv3config-supply-custom-field_table--At1mxQWSBpccxtSzp8aJXnOtAg-qnsdEpUuSPoimmslm1bPN) that @mcgibbon outlined earlier after a discussion with me and @oliverwm1.

I definitely see value in automatically setting the field_table used to prevent user-error as much as possible, so eventually I think we could look to somehow support the SHiELD use-case from that perspective; however, that introduces a few more complications that we'd have to think through:
- The microphysics scheme used is set with different namelist parameters than `imp_physics`.
- The new boundary layer scheme requires a new tracer for TKE, though one can still run with older PBL schemes, which do not require this additional tracer (so the field_table not only is dependent on the microphysics scheme, but also is dependent on the boundary layer scheme).

I don't think there's a particular downside to allowing users (at their own risk) to explicitly set the field_table used, so hopefully we're all on board for the spirit of this change!

Thanks @mcgibbon for pointing me to the `tempfile` module; that was quite useful for writing these tests.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -81,29 +81,22 @@ filesystem must be given as absolute paths. If paths are given that begin with `
 attempt to download the specified file or files from Google Cloud Storage. For this functionality, ``gcsfs``
 must be installed and authorized to download from the specified bucket.
 
-A ``field_table`` file is required for running the model, though it is not
-necessarily a required key in the ``config`` dictionary. In this file, names
-and properties of the different tracers used in the model are specified. If
-not provided in the ``config`` dictionary, ``fv3config`` will attempt to infer
-the required ``field_table`` based on the namelist parameters specified. This
-is typically recommended, because different ``field_table`` files are required
-depending on the model configuration, and it can be nice to rely on
-``fv3config`` to decide on what the appropriate file should be. That said, it
-can be useful to be able to override this. If one specifies a file path as a
-``field_table``, this file will be used regardless of the namelist
-parameters. If one specifies a directory as a ``field_table`` it will be
-assumed that the reference ``field_table`` files will be present in this
-directory, and will be used in place of the default reference ``field_table``
-files provided with ``fv3config``.  The current reference ``field_table`` files
-provided in ``fv3config`` correspond with files used for running with Zhao-Carr
-microphysics (``field_table_ZhaoCarr``) or with GFDL microphysics
-(``field_table_GFDLMP``). In FV3GFS these correspond with running with
-``gfs_phys_nml`` namelist parameters of ``imp_physics=99``, ``ncld=1``, and
-``imp_physics=11``, ``ncld=5``, respectively.
-
 The ``namelist`` item is special in that it is explicitly stored in the ``config`` dictionary. For the
 fv3gfs model, individual namelists are specified for various components of the model. As an example, the
 vertical resolution can be accessed via ``config['namelist']['fv_core_nml']['npz']``.
+
+By default, fv3config attempts to automatically select the ``field_table`` file
+to use for the model based on the selected microphysics scheme in the
+namelist. This supports Zhao-Carr or GFDL microphysics. If the user provides a
+``field_table`` key indicating a filename in the configuration dictionary, that
+file will be used instead.
+
+.. note::
+   The `Han and Bretherton (2019) <https://journals.ametsoc.org/doi/full/10.1175/WAF-D-18-0146.1>`_ TKE-EDMF
+   boundary layer scheme requires an additional tracer to be defined in the
+   ``field_table`` for TKE. This scheme is currently not supported by default
+   in ``fv3config``; however for the time being one can supply a custom
+   ``field_table`` for this purpose.
 
 Some helper functions exist for editing and retrieving information from configuration
 dictionaries, like :py:func:`fv3config.get_run_duration` and
@@ -266,3 +259,5 @@ restart initial conditions can be created with::
     config['initial_conditions'] = 'restart_example'
     config = enable_restart(config)
     write_run_directory(config, './rundir')
+
+.. _HB2019: https://journals.ametsoc.org/doi/full/10.1175/WAF-D-18-0146.1
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -81,6 +81,26 @@ filesystem must be given as absolute paths. If paths are given that begin with `
 attempt to download the specified file or files from Google Cloud Storage. For this functionality, ``gcsfs``
 must be installed and authorized to download from the specified bucket.
 
+A ``field_table`` file is required for running the model, though it is not
+necessarily a required key in the ``config`` dictionary. In this file, names
+and properties of the different tracers used in the model are specified. If
+not provided in the ``config`` dictionary, ``fv3config`` will attempt to infer
+the required ``field_table`` based on the namelist parameters specified. This
+is typically recommended, because different ``field_table`` files are required
+depending on the model configuration, and it can be nice to rely on
+``fv3config`` to decide on what the appropriate file should be. That said, it
+can be useful to be able to override this. If one specifies a file path as a
+``field_table``, this file will be used regardless of the namelist
+parameters. If one specifies a directory as a ``field_table`` it will be
+assumed that the reference ``field_table`` files will be present in this
+directory, and will be used in place of the default reference ``field_table``
+files provided with ``fv3config``.  The current reference ``field_table`` files
+provided in ``fv3config`` correspond with files used for running with Zhao-Carr
+microphysics (``field_table_ZhaoCarr``) or with GFDL microphysics
+(``field_table_GFDLMP``). In FV3GFS these correspond with running with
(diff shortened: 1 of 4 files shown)


docs/usage.rst
```diff
@@ -81,6 +81,26 @@ filesystem must be given as absolute paths. If paths are given that begin with `
...
+``field_table``, this file will be used regardless of the namelist
+parameters. If one specifies a directory as a ``field_table`` it will be
+assumed that the reference ``field_table`` files will be present in this
+directory, and will be used in place of the default reference ``field_table``
+files provided with ``fv3config``.  The current reference ``field_table`` files
+provided in ``fv3config`` correspond with files used for running with Zhao-Carr
+microphysics (``field_table_ZhaoCarr``) or with GFDL microphysics
+(``field_table_GFDLMP``). In FV3GFS these correspond with running with
+``gfs_phys_nml`` namelist parameters of ``imp_physics=99``, ``ncld=1``, and
+``imp_physics=11``, ``ncld=5``, respectively.
```
mcgibbon: Very descriptive, but so much text! Can I suggest something shorter?

> By default, fv3config attempts to automatically select the ``field_table`` file to use for the model based on the selected microphysics scheme in the namelist. This supports Zhao-Carr or GFDL microphysics. If the user provides a ``field_table`` key indicating a filename in the configuration dictionary, that file will be used instead.

If you want to come up with a different shorter paragraph feel free. If you think some of the extra "detail" is important you can add it as a [note/warning](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html). Also, this bit should probably go after the `namelist` is introduced in the paragraph below.

fv3config/_datastore.py
```diff
@@ -181,6 +181,45 @@ def get_microphysics_name(config):
...
+        str: field_table filename
+
+    Raises:
+        ConfigError
+    """
+    field_table = config.get("field_table", None)
+    if field_table is None:
+        field_table_dir = os.path.join(DATA_DIR, "field_table")
+    elif filesystem.isabs(field_table) and filesystem.get_fs(field_table).exists(
+        field_table
```
mcgibbon: Don't check `exists` here. If the field table does not exist, we want the program to error and tell the user they've given an invalid option, instead of silently failing and selecting the default option.

mcgibbon: Should add a test on this function that makes sure it raises `ConfigError` (in the else block below) if it's given a non-existent absolute path.

fv3config/_datastore.py
```diff
@@ -181,6 +181,45 @@ def get_microphysics_name(config):
 def get_field_table_filename(config):
     """Get field_table filename given configuration dictionary
```
mcgibbon: This function is a little large now, with very low-level code, can it be refactored to call only higher-level functions that describe what they do based on their name, that contain the low-level code? And can it also be refactored to only return at the bottom of the function? It's hard to follow the control-flow.

mcgibbon: I should add it's not your fault it's like this, it was already getting there before you added this new feature, which pushed it over the top.

tests/test_asset_list.py
```diff
@@ -147,6 +148,42 @@ def test_get_field_table_asset_default(self):
...
+            expected_field_table_asset = get_asset_dict(
+                directory, filename, target_name="field_table"
+            )
+            self.assertEqual(field_table_asset, expected_field_table_asset)
+
+    def test_get_field_table_asset_non_existent_filename_or_directory(self):
+        config = fv3config.get_default_config()
+        config["field_table"] = "foo"
+        with self.assertRaises(fv3config.ConfigError):
+            get_field_table_asset(config)
```
mcgibbon: Add a case where the non-existent filename is a non-existent absolute path (e.g. `/foo`). I think right now that test would fail. Apart from that distinction you covered all the cases, great job!

tests/test_asset_list.py
```diff
@@ -147,6 +148,42 @@ def test_get_field_table_asset_default(self):
...
+        with tempfile.NamedTemporaryFile() as field_table:
+            config["field_table"] = field_table.name
+            field_table_asset = get_field_table_asset(config)
+            directory, filename = os.path.split(field_table.name)
+            expected_field_table_asset = get_asset_dict(
+                directory, filename, target_name="field_table"
+            )
+            self.assertEqual(field_table_asset, expected_field_table_asset)
+
+    def test_get_field_table_asset_non_existent_filename_or_directory(self):
```
mcgibbon: replace "filename_or_directory" with "path"

fv3config/_datastore.py
spencerkclark: Happy to add another test; I think this behaves as desired in that case, however, unless I'm not understanding.  Agree with your earlier comment though that this function could be broken up to make things clearer!

spencerkclark:
Many thanks for the review @mcgibbon -- all good comments.  I'll push some updates tomorrow morning.

docs/usage.rst
spencerkclark: Thanks for the suggested text.  I agree that's a nicer, more concise way of putting things.  I did add a note regarding the TKE-EDMF scheme, because I do think that's important for folks to be aware of, particularly when switching to the new fortran repo, which supports the new scheme, and is the scheme SHiELD is currently being run with at GFDL.

fv3config/_datastore.py
spencerkclark: I refactored things a bit; let me know what you think.

fv3config/_datastore.py
```diff
@@ -187,6 +207,35 @@ def get_field_table_filename(config):
...
+    field_table = config.get("field_table", None)
+    if field_table is None:
+        directory = os.path.join(DATA_DIR, "field_table")
+        return get_inferred_field_table_filename(config, directory)
+    else:
+        if not filesystem.is_existing_absolute_path(field_table):
+            raise ConfigError(
+                f"field_table={field_table} must either be left unset or set "
+                "to an existing absolute path to a file or directory"
+            )
```
mcgibbon: We talked on the phone briefly about combining the three checks of this into a function call that occurs only once at the end on a field_table path. You will also have to check if the given `field_table` is a directory, but it won't be necessary to raise if it doesn't exist because the file will also not exist in the check at the end.

fv3config/filesystem.py
```diff
@@ -29,6 +29,11 @@ def isabs(path: str) -> bool:
         return os.path.isabs(path)
 
 
+def is_existing_absolute_path(path: str) -> bool:
+    """Return whether the path is an existing absolute path"""
+    return isabs(path) and get_fs(path).exists(path)
```
mcgibbon: Nice and concise!

tests/test_asset_list.py
```diff
@@ -147,6 +148,48 @@ def test_get_field_table_asset_default(self):
...
+        config = fv3config.get_default_config()
+        config["field_table"] = "foo"
+        with self.assertRaises(fv3config.ConfigError):
+            get_field_table_asset(config)
+
+    def test_get_field_table_asset_non_existent_absolute_path(self):
+        config = fv3config.get_default_config()
+        config["field_table"] = "/foo"
+        with self.assertRaises(fv3config.ConfigError):
+            get_field_table_asset(config)
```
mcgibbon: Thanks!

docs/usage.rst
```diff
@@ -85,6 +85,19 @@ The ``namelist`` item is special in that it is explicitly stored in the ``config
...
+namelist. This supports Zhao-Carr or GFDL microphysics. If the user provides a
+``field_table`` key indicating a filename in the configuration dictionary, that
+file will be used instead.
+
+.. note::
+   The `Han and Bretherton (2019) <https://journals.ametsoc.org/doi/full/10.1175/WAF-D-18-0146.1>`_ TKE-EDMF
+   boundary layer scheme requires an additional tracer to be defined in the
+   ``field_table`` for TKE. This scheme is currently not supported by default
+   in ``fv3config``; however for the time being one can supply a custom
+   ``field_table`` for this purpose.
```
mcgibbon: Great! I like the citation.

fv3config/_datastore.py
spencerkclark: Thanks @mcgibbon!  I think I figured out a decent solution.  Please have another quick look.

fv3config/_datastore.py
```diff
@@ -187,6 +198,33 @@ def get_field_table_filename(config):
     Returns:
         str: field_table filename
 
+    Raises:
+        ConfigError
+    """
+    field_table = config.get("field_table", DEFAULT_FIELD_TABLE_DIR)
+    field_table_filename = _return_or_infer_field_table_filename(config, field_table)
+
```
mcgibbon: Can delete this whitespace.

PR #63, opened 2020-02-28
orographic_forcing can optionally be a cloud bucket

rheacangeo:
Enable setting an 'orographic_forcing' in the configuration yaml that can be a cloud bucket, while if this key missing or set to default will use the local data cache as it did before.

### DIFF
--- a/fv3config/_datastore.py
+++ b/fv3config/_datastore.py
@@ -20,6 +20,10 @@ FORCING_OPTIONS_DICT = {
     "default": "base_forcing",
 }
 
+OROGRAPHIC_FORCING_OPTIONS_DICT = {
+    "default": "orographic_data",
+}
+
 INITIAL_CONDITIONS_OPTIONS_DICT = {
     "gfs_example": "initial_conditions/gfs_initial_conditions",
     "restart_example": "initial_conditions/restart_initial_conditions",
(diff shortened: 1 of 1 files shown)


fv3config/_datastore.py
```diff
@@ -45,11 +49,15 @@ def get_orographic_forcing_directory(config):
...
-    dirname = os.path.join(get_internal_cache_dir(), f"orographic_data/{resolution}")
-    if not os.path.isdir(dirname):
-        valid_options = os.listdir(
-            os.path.join(get_internal_cache_dir(), "orographic_data")
-        )
+    if 'orographic_forcing' not in config:
+        orographic_forcing_spec = 'default'
+    else:
+        orographic_forcing_spec = config['orographic_forcing']
+    parent_dirname = resolve_option(orographic_forcing_spec, OROGRAPHIC_FORCING_OPTIONS_DICT)
```
mcgibbon: `orographic_forcing_spec` isn't any more descriptive than `config.get('orographic_forcing', 'default')`, so I'd suggest replacing this with:
`parent_dirname = resolve_option(config.get('orographic_forcing', 'default'), OROGRAPHIC_FORCING_OPTIONS_DICT)`

fv3config/_datastore.py
```diff
@@ -45,11 +49,15 @@ def get_orographic_forcing_directory(config):
...
+    dirname = os.path.join(parent_dirname, resolution)
+
+    if get_internal_cache_dir() in dirname and not os.path.isdir(dirname):
+        valid_options = os.listdir(parent_dirname)
```
mcgibbon: This should behave the same whether it's local or remote - remote stores are expected to look the same as the internal cache, with resolution subdirectories. You can perform these `os` operations in a way that is also compatible with remote paths by doing `fs = filesystem.get_fs(dirname)` followed by `fs.isdir(dirname)` and `fs.listdir(parent_dirname)`. The one `os` method here that can be used on anything is `os.path.join`.

fv3config/_datastore.py
rheacangeo: sure

PR #66, opened 2020-03-19
Changes needed for the fv3net one-step jobs

nbren12:
The PR changes the behavior of `run_kubernetes` to not ensure that the locations of the fv3config.yml and the output directory are remote. This assumption is too stringent. run_k8s should be able to use fv3configs stored in the container's file system and save to arbitrary locations (including ones in the container's filesystems or k8s local volume that is mounted).

Also fix a bug when generating the container name for docker images with an "_" in them. K8s does not allow "_" in the name of container.

### DIFF
--- a/fv3config/fv3run/_kubernetes.py
+++ b/fv3config/fv3run/_kubernetes.py
@@ -2,7 +2,6 @@ import os
 import re
 import uuid
 from .._exceptions import DelayedImportError
-from .. import filesystem
 from ._docker import _get_python_command
 
 try:
--- a/fv3config/fv3run/_kubernetes.py
+++ b/fv3config/fv3run/_kubernetes.py
@@ -89,7 +89,6 @@ def _get_job(
     image_pull_policy="IfNotPresent",
     job_labels=None,
 ):
-    # _ensure_locations_are_remote(config_location, outdir)
     kube_config = KubernetesConfig(
         jobname,
         memory_gb,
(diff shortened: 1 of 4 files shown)


tests/test_kubernetes.py
```diff
@@ -69,20 +69,6 @@ def job_labels(request):
...
-        )
-
-
-def test_local_outdir_rejected(config_location, docker_image, runfile):
-    with pytest.raises(ValueError):
-        fv3config.fv3run._kubernetes._get_job(
-            config_location, "/local/outdir", docker_image, runfile
-        )
-
-
```
mcgibbon: Can you replace these with tests that they *are not* rejected?

fv3config/fv3run/_kubernetes.py
```diff
@@ -89,7 +88,6 @@ def _get_job(
     image_pull_policy="IfNotPresent",
     job_labels=None,
 ):
-    _ensure_locations_are_remote(config_location, outdir)
```
mcgibbon: Can you print some kind of warning if outdir is remote? Since presently it basically means "actually set no outdir". Alternatively allow outdir=None or remote, and if it's None use a specific local path in /var/tmp or something.

nbren12: added a warning.

tests/test_kubernetes.py
nbren12: Hmm. It does seem like the reason for these tests no longer exists. I added a local path to `outdir` fixture so that option should be picked up by other tests.

mcgibbon: OK, thanks. I hadn't noticed you added the option to the outdir fixture.

fv3config/fv3run/_kubernetes.py
```diff
@@ -89,7 +91,12 @@ def _get_job(
...
+    if filesystem._is_local_path(outdir):
+        warnings.warn(
+            f"Output directory {outdir} is a local path, so it will not be accesible "
```
mcgibbon: typo in "accessible"

PR #68, opened 2020-03-20
Pass arbitrary kwargs from run_docker/run_k8s to run_native

nbren12:
Changing the signature of run_native (e.g. to add an option changing the logging behavior), requires editing 3 other files: `run_kubernetes`, `run_docker`, and the `__main__.py`. This PR allows `run_kubernetes` and `run_docker` to pass arbitrary arguments to `run_native` provided that they are serializable with `json`. A side effect is that `run_kubernetes` can now take an actual configuration dictionary rather than a url pointing to one. 

It also tidies up the implementation of `run_kubernetes` especially. `run_docker` still has some unavoidable logic to do with creating the bind-mounts, but is slightly cleaner now too.

### DIFF
--- a/fv3config/fv3run/_docker.py
+++ b/fv3config/fv3run/_docker.py
@@ -38,8 +38,7 @@ def run_docker(
         keyfile = os.environ.get("GOOGLE_APPLICATION_CREDENTIALS", None)
 
     if isinstance(config_dict_or_location, str):
-        with fsspec.open(config_dict_or_location) as f:
-            config_dict = yaml.safe_load(f.read())
+        config_dict = _load_yaml(config_dict_or_location)
     else:
         config_dict = config_dict_or_location
 
@@ -49,16 +48,21 @@ def run_docker(
     _get_credentials_args(keyfile, docker_args, bind_mount_args)
     _get_local_data_bind_mounts(config_dict, bind_mount_args)
 
-    outdir_in_docker = _get_docker_args(docker_args, bind_mount_args, outdir)
+    _get_docker_args(docker_args, bind_mount_args, outdir)
     runfile_in_docker = _get_runfile_args(runfile, bind_mount_args)
 
     python_command = run_native.command(
-        config_dict, outdir_in_docker, runfile=runfile_in_docker, **kwargs
+        config_dict, DOCKER_OUTDIR, runfile=runfile_in_docker, **kwargs
     )
 
     subprocess.check_call(
         DOCKER_COMMAND + bind_mount_args + docker_args + [docker_image] + python_command
     )
+
+
+def _load_yaml(url):
+    with fsspec.open(url) as f:
+        return yaml.safe_load(f.read())
 
 
 def _get_runfile_args(runfile, bind_mount_args) -> str:
@@ -123,7 +127,6 @@ def _get_local_data_bind_mounts(config_dict, bind_mount_args):
 def _get_docker_args(docker_args, bind_mount_args, outdir):
     bind_mount_args += ["-v", f"{os.path.abspath(outdir)}:{DOCKER_OUTDIR}"]
     docker_args += ["--rm", "--user", f"{os.getuid()}:{os.getgid()}"]
-    return DOCKER_OUTDIR
 
 
 def _get_credentials_args(keyfile, docker_args, bind_mount_args):
--- a/fv3config/fv3run/_docker.py
+++ b/fv3config/fv3run/_docker.py
@@ -54,7 +54,7 @@ def run_docker(
     outdir_in_docker = _get_outdir_args(docker_args, bind_mount_args, outdir)
     runfile_in_docker = _get_runfile_args(runfile, bind_mount_args)
 
-    python_command = run_native.command(config_dict, outdir_in_docker, runfile=runfile_in_docker)
+    python_command = run_native.command(config_dict, outdir_in_docker, runfile=runfile_in_docker, **kwargs)
 
     subprocess.check_call(
         DOCKER_COMMAND
--- a/fv3config/fv3run/_docker.py
+++ b/fv3config/fv3run/_docker.py
@@ -1,10 +1,9 @@
-import tempfile
 import subprocess
 import os
 import fsspec
 import yaml
 from .. import filesystem
-from ._native import CONFIG_OUT_FILENAME, _get_config_dict_and_write, run_native
+from ._native import CONFIG_OUT_FILENAME, run_native
 
 DOCKER_OUTDIR = "/outdir"
 DOCKER_CONFIG_LOCATION = os.path.join("/", CONFIG_OUT_FILENAME)
@@ -15,8 +14,7 @@ FV3RUN_MODULE = "fv3config.fv3run"
 
 
 def run_docker(
-    config_dict_or_location, outdir, docker_image, runfile=None, keyfile=None,
-    **kwargs
+    config_dict_or_location, outdir, docker_image, runfile=None, keyfile=None, **kwargs
 ):
     """Run the FV3GFS model in a docker container with the given configuration.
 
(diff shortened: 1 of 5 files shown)


fv3config/fv3run/_docker.py
```diff
@@ -35,59 +36,38 @@ def run_docker(
...
     if keyfile is None:
         keyfile = os.environ.get("GOOGLE_APPLICATION_CREDENTIALS", None)
-    filesystem.get_fs(outdir).makedirs(outdir, exist_ok=True)
-    with tempfile.NamedTemporaryFile(suffix=".yaml") as config_tempfile:
-        bind_mount_args = []
-        python_args = []
-        docker_args = []
-        config_location = _get_config_args(
-            config_dict_or_location, config_tempfile, bind_mount_args
-        )
```
mcgibbon: I can't find where config_location is being added to bind_mount_args now, can you point me to it or add it?

fv3config/fv3run/_docker.py
```diff
@@ -143,6 +123,7 @@ def _get_local_data_bind_mounts(config_dict, bind_mount_args):
 def _get_docker_args(docker_args, bind_mount_args, outdir):
     bind_mount_args += ["-v", f"{os.path.abspath(outdir)}:{DOCKER_OUTDIR}"]
     docker_args += ["--rm", "--user", f"{os.getuid()}:{os.getgid()}"]
+    return DOCKER_OUTDIR
```
mcgibbon: You shouldn't need to return a global constant in a module-private function.

fv3config/fv3run/_docker.py
```diff
@@ -35,59 +36,38 @@ def run_docker(
...
 
-def _get_python_command(config_location, outdir):
-    return ["python3", "-m", FV3RUN_MODULE, config_location, outdir]
+    filesystem.get_fs(outdir).makedirs(outdir, exist_ok=True)
+    bind_mount_args = []
+    docker_args = []
+    _get_credentials_args(keyfile, docker_args, bind_mount_args)
+    _get_local_data_bind_mounts(config_dict, bind_mount_args)
 
+    outdir_in_docker = _get_docker_args(docker_args, bind_mount_args, outdir)
```
mcgibbon: should just use DOCKER_OUTDIR in place of this local variable (which will currently always be equal to DOCKER_OUTDIR) wherever you would use it.

fv3config/fv3run/_kubernetes.py
```diff
@@ -5,7 +5,7 @@
 
 from .. import filesystem
 from .._exceptions import DelayedImportError
-from ._docker import _get_python_command
+from ._native import run_native
```
mcgibbon: Positive change, thanks!

fv3config/fv3run/_kubernetes.py
```diff
@@ -26,6 +26,8 @@ def run_kubernetes(
     gcp_secret=None,
     image_pull_policy="IfNotPresent",
     job_labels=None,
+    submit=True,
+    **kwargs,
```
mcgibbon: As we discussed, in the short term the new kwarg should be listed here explicitly. In the longer term we should get a kubernetes submit command that takes in a command string, and get that command string from another function.

fv3config/fv3run/_kubernetes.py
```diff
@@ -61,12 +63,18 @@ def run_kubernetes(
...
         job_labels (Mapping[str, str], optional): labels provided as key-value pairs
             to apply to job pod.  Useful for grouping jobs together in status checks.
+        **kwargs:  other arguments passed to be passed to run_native. These must be serializeable with json.
     """
+
+    if filesystem._is_local_path(outdir):
+        warnings.warn(
+            f"Output directory {outdir} is a local path, so it will not be accessible "
+            "once the job finishes."
+        )
```
mcgibbon: This warning might not be possible with a command-centric approach, though we could put it instead in the function that generates the command? Or maybe we'll keep this API the same and it will just make use of the new interface. idk.

fv3config/fv3run/_kubernetes.py
```diff
@@ -117,12 +119,12 @@ def _submit_job(job, namespace):
     api.create_namespaced_job(body=job, namespace=namespace)
 
 
-def _create_job_object(config_location, outdir, docker_image, runfile, kube_config):
+def _create_job_object(command, docker_image, kube_config):
```
mcgibbon: Good!

fv3config/fv3run/_native.py
```diff
@@ -19,7 +21,28 @@
...
+        args, kwargs = json.loads(argv[1])
+        func(*args, **kwargs)
+
+    def command(*args, **kwargs) -> str:
+        serialized = json.dumps([args, kwargs])
+        return ["python", "-m", this_module, serialized]
+
+    func.main = main
+    func.command = command
+    return func
```
mcgibbon: Neato

fv3config/fv3run/_native.py
```diff
@@ -19,7 +21,28 @@
...
+def call_via_subprocess(func):
+    this_module = func.__module__
+
+    def main(argv):
+        import json
+
+        args, kwargs = json.loads(argv[1])
+        func(*args, **kwargs)
+
+    def command(*args, **kwargs) -> str:
```
mcgibbon: can use `@functools.wraps` to give this function the same call signature as `run_native`. You may want to figure out how to just modify the call signature without modifying the name or docstring, and see if there are any side-effects I don't know about.

tests/test_fv3run.py
```diff
@@ -454,3 +414,42 @@ def test_get_credentials_args(keyfile, expected_docker_args, expected_bind_mount
...
+    def dummy_function(*obj):
+        output.append(hash(obj))
+
+    inputs = (1, 2, 3)
+
+    # first call is via "main"
+    serialized_args = dummy_function.command(*inputs)[-1]
+    argv = [None, serialized_args]
+    dummy_function.main(argv)
+    dummy_ans = output.pop()
```
mcgibbon: If you do add the code that modifies the signature of `command` to be the same as the wrapped function, be sure to test it :)

fv3config/fv3run/_native.py
nbren12: I added the decorator. However, after reading the [docs](https://docs.python.org/2/library/functools.html#functools.update_wrapper), I don't think `functools.wraps` does any sort of argument checking. Do you think that feature is important to include, since atm this is not a public API, and we can just assume that run_k8s and run_docker know what they are doing?

tests/test_fv3run.py
nbren12: I would rather leave this safety guarantee to another PR unless you think it is very important.

fv3config/fv3run/_kubernetes.py
mcgibbon: OK, this was mostly a note about an issue you'll run into in the future - you'll need a special command for submitting kubernetes fv3gfs runs basically so that you can post this warning.

tests/test_fv3run.py
```diff
@@ -223,43 +217,6 @@ def maybe_get_file(*args, **kwargs):
         pass
 
 
-@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
mcgibbon: OK

tests/test_kubernetes.py
```diff
@@ -151,13 +153,6 @@ def _check_toleration(toleration):
     assert toleration.value == "climate-sim-pool"
 
 
-def _check_command(command, outdir, config_location, runfile):
```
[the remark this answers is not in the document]
mcgibbon: OK, as long as it's tested.

tests/test_fv3run.py
mcgibbon: It's an easy test. just something like `with pytest.raises(TypeError): command(bad=None, args=True)`. I think it's important enough to include, as it can allow weird, hard-to-track bugs if this regresses in a way that isn't intentional.

nbren12:
@mcgibbon FYI, functools wraps doesn't enforce that the signatures match:
```
>>> from functools import wraps
>>> def f(a,b):
...     pass
...
>>> @wraps(f)
... def g(a):
...     pass
...
>>> g(10)
```

fv3config/fv3run/_native.py
```diff
@@ -19,7 +23,34 @@
 logger = logging.getLogger("fv3run")
 
 
-def run_native(config_dict_or_location, outdir, runfile=None):
+def call_via_subprocess(func):
+    this_module = func.__module__
+    signature = inspect.signature(func)
+
+    def main(argv):
+        import json
```
mcgibbon: import not needed here

PR #72, opened 2020-03-25
Allow enabling and disabling caching

mcgibbon:
Major changes:
- Adds a routine `fv3config.do_remote_caching` which sets the default caching behavior for remote files.

Minor changes:
- Renamed the `cache_location` module to `caching`.

### DIFF
--- a/fv3config/caching.py
+++ b/fv3config/caching.py
@@ -8,6 +8,15 @@ else:
     os.makedirs(USER_CACHE_DIR, exist_ok=True)
 CACHE_PREFIX = "fv3config-cache"
 
+CACHE_REMOTE_FILES = True
+
+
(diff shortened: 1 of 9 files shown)


fv3config/caching.py
```diff
@@ -8,6 +8,15 @@
     os.makedirs(USER_CACHE_DIR, exist_ok=True)
 CACHE_PREFIX = "fv3config-cache"
 
+CACHE_REMOTE_FILES = True
+
+
+def do_remote_caching(flag: bool):
```
oliverwm1: docstring? short one would be helpful (and maybe a good place to state that default is True?)

mcgibbon: Added.

PR #73, opened 2020-03-30
Add btl_vader_single_copy_mechanism MPI flag to fv3run

mcgibbon:
This PR adds the flag `--mca btl_vader_single_copy_mechanism none` to mpirun in fv3run. It prevents a certain "error" message from being printed repeatedly when the model runs. The error does not affect output, and currently appears so often it is difficult to read the log file.


mcgibbon:
No, I do not know. It seemed to me that the only change was the "error" message doesn't print when I was trying with and without in September.

PR #75, opened 2020-04-02
Add --kubernetes flag and run model in outdir when outdir is local

mcgibbon:
Major changes:
- Added `--kubernetes` command-line flag to output a kubernetes config yaml to stdout
- When output location is set to a local path, the job now runs in that output location instead of in a temporary directory which then gets copied. This is done both to reduce copying time for large jobs, and to improve visibility of model behavior while running.

Minor changes:
- Updated `filesystem.is_local_path` naming to be package-private instead of module-private
- Fixed an issue where the subprocess-wrapped run_native lived inside an imported module, resulting in python warnings

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,27 +5,32 @@ History
 latest
 ------
 
+Major changes:
+~~~~~~~~~~~~~
+- Added `--kubernetes` command-line flag to output a kubernetes config yaml to stdout
+
 Minor changes:
 ~~~~~~~~~~~~~
-* Added the flag ``--mca btl_vader_single_copy_mechanism none to mpirun in fv3run`` to mpirun in fv3run
-* Add ReadTheDocs configuration file
-* Do not require output dir and fv3config to be remote in ``run_kubernetes``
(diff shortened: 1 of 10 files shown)


oliverwm1:
LGTM. Can you add the `--kubernetes` option to the docs?

Is there a plan to allow more precise specification of Kubernetes resources from the command line? The feature as it currently stands seems of somewhat limited utility if it only allows runs with cpu_count=1 and memory_gb=3.6GB, the current defaults for `run_kubernetes`.

nbren12:
@oliverwm1 I think it would be fair for the `fv3run` CLI to support all the current parameters of `run_kuberenetes`. That said, @mcgibbon and I are concerned that adding more features to functions like `run_kuberentes` (e.g. volume mounts etc) could result in a function with so many arguments that it basically equivalent to a k8s yaml. At that point, I think we should just write raw K8s jobs as yaml files, templates for which could be generated with run_k8s.

mcgibbon:
I'm modifying this so `--kubernetes` will output the job yml to stdout instead. Will get in touch when it's ready for re-review.

PR #77, opened 2020-04-07
Make output capturing an optional behavior

nbren12:
All of our runtime environments (the linux shell, docker, and kubernetes) have powerful tools for manipulating the stdout and stderr streams, but fv3run captures these streams and instead puts them in relatively inaccessible files. This makes real time debugging and monitoring of k8s jobs especially challenging. In general, it is more flexible to pipe the outputs to stderr and stdout and let the user leverage the power of linux shell, pipes, and commands like `tee` to put the outputs where they want. 

As an example

    fv3run fv3config.yml output 2> output/stderr.log > output/stdout.log

will emulate the previous behavior provided the output directory is made
first.

Changes

* Avoid redirecting stderr and stdout when using the `fv3run` CLI unless the `--capture-output` argument is passed to `run_native`. 
* add an optional argument `capture_output` to `run_native`. Unlike the CLI, it's default value is True to avoid breaking all the tests.
* Remove the `run` function in `__main__`. This does not appear to be public function, and it seems like an unnecessary layer of indirection between the argument parsing and the calls to `run_native`, etc.

### DIFF
--- a/fv3config/fv3run/_native.py
+++ b/fv3config/fv3run/_native.py
@@ -83,13 +83,15 @@ def run_native(
             filesystem.get_file(
                 runfile, os.path.join(localdir, os.path.basename(runfile))
             )
-        with _log_exceptions(localdir):
+        with _error_context(localdir, capture_output) as (stdout, stderr):
             n_processes = get_n_processes(config_dict)
             _run_experiment(
                 localdir,
                 n_processes,
                 runfile=runfile,
                 mpi_flags=_add_oversubscribe_if_necessary(MPI_FLAGS, n_processes),
+                stdout=stdout,
+                stderr=stderr
             )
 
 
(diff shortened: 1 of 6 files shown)


fv3config/fv3run/_native.py
```diff
@@ -133,20 +138,38 @@ def _temporary_directory(outdir):
...
-            "Experiment failed. " "Check %s and %s for logs.",
-            STDOUT_FILENAME,
-            STDERR_FILENAME,
-        )
+        logger.critical("Experiment failed")
         raise e
 
 
+@contextlib.contextmanager
+def _error_context(localdir: str, capture_output: bool):
```
mcgibbon, line 165: I'd perhaps call this "output context" since it redirects output moreso than it does error handling. It doesn't only redirect stderr.

nbren12:
I tested this inside of a docker image so it ran all of the integration tests skipped by the CI. It does print the stderr and stdout if wanted.

PR #80, opened 2020-04-09
Ignore expected warning during kubernetes get_job test

mcgibbon:
This PR adds an ignore flag to the warning that the kubernetes output directory is a local path and that output data will not be available, since the use of a local path is by design in this test (and we're not trying to get output data). This PR returns the fv3config tests to a state where no warnings should be printed when they run.

I'd consider this change too small to document in the HISTORY.rst.

### DIFF
--- a/tests/test_kubernetes.py
+++ b/tests/test_kubernetes.py
@@ -69,6 +69,7 @@ def job_labels(request):
     return request.param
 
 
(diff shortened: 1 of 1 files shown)


mcgibbon:
It could save someone a lot of trouble at some point, we should revisit removing it at some point in the future if it shows up a lot in cases where it doesn't matter.

PR #81, opened 2020-04-15
Feature/parallel upload

mcgibbon:
Major changes:
- filesystem operations now manually backoff with a 1-minute max time on RuntimeError (which gcsfs often raises when it shouldn't) and gcsfs.utils.HttpError
- `put_directory` now makes use of a thread pool to copy items in parallel.

Minor changes:
- `run_docker` now works when supplying an outdir on google cloud storage
- `put_directory` is now marked as package-private instead of module-private

### DIFF
--- a/fv3config/filesystem.py
+++ b/fv3config/filesystem.py
@@ -3,11 +3,12 @@ import fsspec
 import backoff
 from ._exceptions import DelayedImportError
 from . import caching
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ThreadPoolExecutor, Executor
 
 
 try:
     import gcsfs
+
     FSSPEC_ERRORS = (RuntimeError, gcsfs.utils.HttpError)
 except ImportError as err:
     gcsfs = DelayedImportError(err)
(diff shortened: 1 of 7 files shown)


fv3config/filesystem.py
```diff
@@ -61,19 +70,31 @@ def is_local_path(location):
...
         source = os.path.join(os.path.abspath(local_source_dir), token)
         dest = os.path.join(dest_dir, token)
         if os.path.isdir(source):
-            fs.makedirs(dest, exist_ok=True)
-            _put_directory(source, dest, fs)
+            fsspec_backoff(fs.makedirs)(dest, exist_ok=True)  # must be blocking call
+            put_directory(source, dest, fs=fs, executor=executor)
         else:
-            fs.put(source, dest)
+            executor.submit(fsspec_backoff(fs.put), source, dest)
```
oliverwm1, line 95: Was there a particular reason to have our own recursive function instead of just calling `fs.put(local_source_dir, dest_dir, recursive=True)`?

Fine to leave as is--and I'm actually not sure if the multi-threading would work if it wasn't written the way it is now.

mcgibbon: I initially wrote here "It allows the option of controlling the threader in the calling context - you can initialize a ThreadPoolExecutor (or any other executor) in the calling context, and pass it in on the initial call." but I misunderstood your question.

When I wrote this initially (I think I wrote it) I didn't know recursive=True was an option. Probably should have used that initially. At this point, to get threaded copy, we'll need our own recursive function (unless fsspec does threading internally, which it doesn't appear that it does, looking at the source).

oliverwm1: Thanks!

PR #83, opened 2020-07-08
Remove old-style data options

mcgibbon:
Major changes:
- the old "default" data options are removed
- get_default_config() is removed, with a placeholder which says it was removed
- ensure_data_is_downloaded is removed, with a placeholder which says it was removed

Our codebase has since a while back transitioned to use remote locations. `get_default_config` and `ensure_data_is_downloaded` are not referenced anywhere in the fv3net repository. Removing this data will significantly reduce the size of our docker images (by about half). We would also have the option of removing the pre-cached data, since it gets cached automatically in contrast to the old-style data which had to be cached manually.

This has follow-on changes:
fv3gfs-fortran: https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/44
fv3gfs-python: https://github.com/VulcanClimateModeling/fv3gfs-python/pull/83


docs/usage.rst
```diff
@@ -67,20 +51,18 @@ Configuration
...
+==================== ======== ============================================
+Key                  Type     Description
+==================== ======== ============================================
+namelist             dict     Model namelist
+experiment_name      str      Name of experiment to use in output
+diag_table           str      location of diag_table file, or one of ("default", "grid_spec", "no_output")
+data_table           str      location of data_table file, or "default"
+initial_conditions   str      location of directory containing initial conditions data
+forcing              str      location of directory containing forcing data
+==================== ======== ============================================
```
oliverwm1, line 63: Need to add the `orographic_forcing` entry.

mcgibbon: Added.

PR #89, opened 2020-07-23
[DSL-54] Add .jenkins folder and regression testing

ofuhrer:
This is driving me crazy... there must be a better way to test this!


ofuhrer:
launch jenkins

ofuhrer:
launch jenkins

ofuhrer:
@twicki Could you review this with high-prio, since we could use it for the other repos as well.

.jenkins/jenkins.sh
```diff
@@ -0,0 +1,90 @@
...
+# check that host (define in machineEnvironment.sh) and slave are consistent
+echo ${host} | grep "${shortslave}" || exitError 1006 ${LINENO} "host does not contain slave"
+
+# act according to action specified
+case "${action}" in
+
+test )
+
+    # check if test script exists
+    script=".jenkins/test.sh"
```
twicki, line 61: should we move directories here to ensure this works from anywhere and not only root?

.jenkins/test.sh
```diff
@@ -0,0 +1,104 @@
...
+    echo "usage: `basename $0` [-h]"
+    echo ""
+    echo "optional arguments:"
+    echo "-h           show this help message and exit"
+}
+
+parseOptions()
+{
+    # process command line options
+    while getopts ":c:dfhl" opt
```
twicki, line 26: lot of unused flags here

.jenkins/test.sh
```diff
@@ -0,0 +1,104 @@
...
+
+# load machine dependent environment
+if [ ! -f ${envloc}/env/env.${host}.sh ] ; then
+    exitError 1202 ${LINENO} "could not find ${envloc}/env/env.${host}.sh"
+fi
+. ${envloc}/env/env.${host}.sh
+
+
+# set root and move to testuite folder
+root=`pwd`
```
twicki, line 69: what is not ran from root? does a `script_path=$(dirname $(which $0))` make sense?

.jenkins/test.sh
```diff
@@ -0,0 +1,104 @@
...
+
+# run tests
+if [ ! -f requirements_dev.txt ] ; then
+    exitError 1205 ${LINENO} "could not find requirements_dev.txt"
+fi
+python3 -m venv venv
+. ./venv/bin/activate
+pip3 install -r requirements_dev.txt
```
twicki, line 77: should we update pip before doing anything else in the venv?

ofuhrer:
launch jenkins

ofuhrer:
launch jenkins

.jenkins/jenkins.sh
```diff
@@ -0,0 +1,90 @@
...
+}
+
+# echo basic setup
+echo "####### executing: $0 $* (PID=$$ HOST=$HOSTNAME TIME=`date '+%D %H:%M:%S'`)"
+
+# start timer
+T="$(date +%s)"
+
+# check sanity of environment
+test -n "$1" || exitError 1001 ${LINENO} "must pass an argument"
```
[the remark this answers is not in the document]
ofuhrer: Fixed.

.jenkins/test.sh
ofuhrer: I've had bad experience with that.

.jenkins/jenkins.sh
```diff
@@ -0,0 +1,98 @@
...
+# get root directory of where jenkins.sh is sitting
+root=`dirname $0`
+
+# act according to action specified
+case "${action}" in
+
+test )
+
+    # check if test script exists
+    script="${root}/test.sh"
```
mcgibbon, line 69: Between this and the definition of `root`, this is hard-coded to only run `test.sh`. Should this be parametrized in some way? For example, adding the `build` and `deploy` clauses even if they do not exist as `sh` files yet? Reading the comment at the top of the file, I'd expect that I can simply add a build/deploy script and call this with build/deploy (or possibly for any command), but as written it only accepts `test`.

Or you could set `script="${root}/${action}.sh"` to allow arbitrary actions to be defined. This will helpfully say the script does not exist if an invalid action is used.

Even better would be to have an `actions` subdirectory, so `script="${root}/actions/${action}.sh"`. This would avoid someone doing something silly like `jenkins.sh jenkins` which infinitely recurses, and also helps make it clearer that `jenkins.sh` is the entrypoint for these actions when looking at the directory structure.

.jenkins/jenkins.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash -f
+
+# This is the master script used to trigger Jenkins actions.
+# The idea of this script is to keep the amount of code in the "Execute shell" field small
+# Exmple syntax:
```
mcgibbon, line 5:
```suggestion
# Example syntax:
```

.jenkins/test.sh
```diff
@@ -0,0 +1,103 @@
+o!/bin/bash -f
```
mcgibbon, line 1: Can you include a short comment about what the arguments (if any) are for this script? It looks like there are none but it's hard for me to tell with my limited bash knowledge.

Also is this a typo?
```suggestion
#!/bin/bash -f
```

I'd really like to see CircleCI use this same test script. Someone could do this in a later PR.

.jenkins/jenkins.sh
ofuhrer: I like your suggestion. Implemented.

.jenkins/test.sh
ofuhrer: Nice catch!

ofuhrer:
Thanks for the review. It does make sense to use the same testing for both CircleCI and Jenkins, would be great if we could do that.

.jenkins/jenkins.sh
ofuhrer: Fixed.

PR #91, opened 2020-08-06
Add a write_run_directory CLI

### LINKED ISSUE #3: Command-line interface
Would be nice to create run directories from the command line. May use [entrypoints](https://amir.rachum.com/blog/2017/07/28/python-entry-points/) for this. Should be able to create a run directory from a yaml file containing the configuration dictionary, at least. After that, maybe we can use click to specify options, or maybe that's more trouble than it's worth.

nbren12:
This commits installs a write_run_directory CLI into the user path
which can be used to write local rundirectories from local or remote
yaml files.

Resolves #3

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -19,6 +19,17 @@ sub-dictionaries. An example C12 configuration dictionary is in the `tests` dire
 
 A run directory based on a configuration can be written using :py:func:`fv3config.write_run_directory`.
 
+Shell Usage
+-----------
+
+This module installs a command line interface `write_run_directory` that can
+be used to write the run directory from a shell. For example, if the file
+`config.yaml` contains a yaml-encoded configuration dictionary
+
+    write_run_directory config.yaml rundir
+
+will write an FV3 run directory to the path `rundir`.
+
 Data Caching
 ------------
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -30,6 +30,7 @@ be used to write the run directory from a shell. For example, if the file
 
 will write an FV3 run directory to the path `rundir`.
 
+This module also installs a command line interface `fv3run`, which is further detailed below.
 Data Caching
 ------------
 
--- a/fv3config/cli.py
+++ b/fv3config/cli.py
@@ -1,0 +1,26 @@
+import argparse
+
+import yaml
+import fsspec
+
+import fv3config
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser("write_run_directory")
+    parser.add_argument(
+        "config", help="URI to fv3config yaml file. Supports any path used by fsspec."
+    )
+    parser.add_argument(
+        "rundir", help="Desired output directory. Must be a local directory"
+    )
+    return parser.parse_args()
+
+
+def write_run_directory():
+    args = _parse_args()
+
+    with fsspec.open(args.config) as f:
+        config = yaml.safe_load(f)
+
+    fv3config.write_run_directory(config, args.rundir)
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -39,7 +39,12 @@
         "Programming Language :: Python :: 3.7",
     ],
     description="FV3Config is used to configure and manipulate run directories for FV3GFS.",
-    entry_points={"console_scripts": ["fv3run=fv3config.fv3run.__main__:main"]},
+    entry_points={
+        "console_scripts": [
+            "fv3run=fv3config.fv3run.__main__:main",
+            "write_run_directory=fv3config.cli:write_run_directory",
```
mcgibbon, line 45: Is this a specific-enough name for a bash command? I'm slightly worried about name clashes, but I'm OK with the name as-is. Maybe `write_fv3_directory` or `fv3dir`?

fv3config/cli.py
```diff
@@ -0,0 +1,26 @@
...
+    return parser.parse_args()
+
+
+def write_run_directory():
+    args = _parse_args()
+
+    with fsspec.open(args.config) as f:
+        config = yaml.safe_load(f)
+
+    fv3config.write_run_directory(config, args.rundir)
```
mcgibbon, line 26: As it is right now, `write_run_directory` does not put the yaml file at `fv3config.yml` in the run directory. It's an outstanding bug to move this yaml-writing into that command and out of the `run_docker`/`run_native` commands. For now you could write the config to that filename here.

setup.py
nbren12: I think it's okay for now. We can always change it if somebody reports a name clash.

fv3config/cli.py
```diff
@@ -0,0 +1,26 @@
...
+import yaml
+import fsspec
+
+import fv3config
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser("write_run_directory")
+    parser.add_argument(
+        "config", help="URI to fv3config yaml file. Supports any path used by fsspec."
```
SpiderMonkey1975, line 12: Will Daint be able to access a remote fv3config.yaml file? Should work if we are running interactively on a login or data-mover node of Piz Daint.

fv3config/cli.py
nbren12: IMO this seems like a bug that should be addressed by write_run_directory since that fix would apply universally.

fv3config/cli.py
nbren12: It should work if GOOGLE_APPLICATION_CREDENTIALS are set correctly, but I think that's a prerequisite for using fv3config overall.

fv3config/cli.py
mcgibbon: You're more than welcome to fix the issue by fixing `write_run_directory` instead, I just wanted to also give you the "faster" option of writing the file here, if you'd like to leave the `write_run_directory` to a future PR or to someone else.

setup.py
mcgibbon: Yeah, [probably not an issue](https://www.google.com/search?q=write_run_directory).

docs/usage.rst
```diff
@@ -19,6 +19,17 @@ sub-dictionaries. An example C12 configuration dictionary is in the `tests` dire
...
+-----------
+
+This module installs a command line interface `write_run_directory` that can
+be used to write the run directory from a shell. For example, if the file
+`config.yaml` contains a yaml-encoded configuration dictionary
+
+    write_run_directory config.yaml rundir
+
+will write an FV3 run directory to the path `rundir`.
+
```
mcgibbon, line 32: In the absence of more time-consuming overhauls of this doc (which has gotten quite long by now), I think this should make a brief mention of the other command-line interface, so it does not give the impression that `write_run_directory` is the only shell usage.

```suggestion
    write_run_directory config.yaml rundir

will write an FV3 run directory to the path `rundir`.

This module also installs a command line interface `fv3run`, which is further detailed below.
```

ofuhrer:
launch jenkins

PR #93, opened 2020-08-07
Add "bytes" asset type for storing data in the fv3config object

nbren12:
Previously this library could not deploy arbitrary assets to remote file
systems without first storing those assets in a remote storage system (e.g. GCS, Kuberentes config map, etc). This is problematic because 
1. the user must manage these collection of external assets AND this configuration yaml itself. For instance, the configuration yaml will not be valid if this external data is moved or deleted.
1. authentication is needed to work with these remote stores.

These difficulties may be worthwhile for large binary assets (e.g. initial condition netCDFs), but it would be convenient to store some small text files in the configuration dictionary itself. Some examples of files we could use this for include prognostic runfiles, diagtables, serialized scikit-learn models, the nudging_file_list used for the reanalysis nudging workflow, etc.

This PR adds a "bytes" asset type that can save an arbitrary list of
bytes to a specified path. Used in conjuction with patch_files, this
will allow the configuration dict to be in principle a
complete/serializable representation of a run-directory.

Public API added:
- `fv3config.get_bytes_asset_dict`

### DIFF
--- a/fv3config/cli.py
+++ b/fv3config/cli.py
@@ -1,0 +1,26 @@
+import argparse
+
+import yaml
+import fsspec
+
+import fv3config
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser("write_run_directory")
+    parser.add_argument(
+        "config", help="URI to fv3config yaml file. Supports any path used by fsspec."
+    )
+    parser.add_argument(
+        "rundir", help="Desired output directory. Must be a local directory"
+    )
+    return parser.parse_args()
+
+
+def write_run_directory():
+    args = _parse_args()
+
+    with fsspec.open(args.config) as f:
+        config = yaml.safe_load(f)
+
+    fv3config.write_run_directory(config, args.rundir)
--- a/tests/test_asset_list.py
+++ b/tests/test_asset_list.py
@@ -303,5 +303,16 @@ def test_write_bytes_asset(tmpdir, target_location):
         assert f.read() == b"hello world"
 
 
+def test_bytes_asset_serializes_with_yaml():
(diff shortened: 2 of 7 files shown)


tests/test_asset_list.py
```diff
@@ -289,5 +290,14 @@ def test_write_asset(self):
         self.assertTrue(os.path.exists(os.path.join(target_workdir, test_filename)))
 
 
+def test_write_bytes_asset(tmpdir):
+    asset = get_bytes_asset_dict(
+        b"hello world", target_location="", target_name="hello.txt"
+    )
+    write_asset(asset, tmpdir)
+    with tmpdir.join("hello.txt").open("rb") as f:
+        assert f.read() == b"hello world"
```
mcgibbon, line 299: Can you add a test where the asset is being written to a subdirectory? Copy-pasting this test and making a minor change is sufficient.

fv3config/cli.py
```diff
@@ -0,0 +1,26 @@
...
+    return parser.parse_args()
+
+
+def write_run_directory():
+    args = _parse_args()
+
+    with fsspec.open(args.config) as f:
+        config = yaml.safe_load(f)
+
+    fv3config.write_run_directory(config, args.rundir)
```
mcgibbon: It feels strange for `fv3config` to define two functions both called `write_run_directory`, could you change the name of this one? You can still keep the same console script name.

nbren12:
Thanks. Yah. I updated the branch and those changes disappeared.

nbren12:
Okay. Good call on the subdirectory test @mcgibbon. That functionality was broken.

PR #94, opened 2020-08-18
Make asset_list_from_path public

brianhenn:
This PR makes asset_list_from_path a public API function. This function does exactly what is needed to get the multiple files that compose a keras model added to the asset list for copying over to the run directory, but it is not currently public.


brianhenn:
Thanks @mcgibbon

PR #95, opened 2020-08-26
Copy configuration dict into rundirectory

nbren12:
Many uses of the python wrapper require a configuration dictionary to be
saved as a yaml file named `fv3config.yml` in the run directory root.

This PR adds this simple capability.

The test is a little tautological, but oh well...

### DIFF
--- a/tests/test_write_run_directory.py
+++ b/tests/test_write_run_directory.py
@@ -137,4 +137,4 @@ class CacheDirectoryTests(unittest.TestCase):
         ] = "gs://vcm-fv3config/config/diag_table/default/v1.0/diag_table"
         with tempfile.TemporaryDirectory() as rundir:
             fv3config.write_run_directory(config, rundir)
-        assert "fv3config.yml" in os.listdir(rundir)
+            assert "fv3config.yml" in os.listdir(rundir)
--- a/tests/test_write_run_directory.py
+++ b/tests/test_write_run_directory.py
@@ -129,3 +129,12 @@ class CacheDirectoryTests(unittest.TestCase):
         with tempfile.TemporaryDirectory() as rundir:
             fv3config.write_run_directory(config, rundir)
         assert os.path.getmtime(cache_filename) == modification_time
+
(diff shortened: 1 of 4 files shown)


mcgibbon:
> I think you also need to modify fv3run so it doesn’t copy the dictionary (what is now a duplicate time). We're planning to remove that code, but it's not removed yet.

Since the tests still pass, it's probably fine to double-copy the dictionary, and you could choose not to do this.

nbren12:
This is ready for re-review @mcgibbon. I added the end to end test and renamed the test_cache.py to clarify.

tests/test_write_run_directory.py
```diff
@@ -129,3 +129,12 @@ def test_cached_diag_table_is_not_redownloaded(self):
         with tempfile.TemporaryDirectory() as rundir:
             fv3config.write_run_directory(config, rundir)
         assert os.path.getmtime(cache_filename) == modification_time
+
+    def test_rundir_contains_fv3config_yml(self):
+        config = copy.deepcopy(DEFAULT_CONFIG)
+        config[
+            "diag_table"
+        ] = "gs://vcm-fv3config/config/diag_table/default/v1.0/diag_table"
+        with tempfile.TemporaryDirectory() as rundir:
```
mcgibbon, line 138: May as well remove the "special" diag table to make the configuration more "standard" (assuming you copied this from a remote diag table test).
```suggestion
        with tempfile.TemporaryDirectory() as rundir:
```

nbren12:
@mcgibbon @ofuhrer What's up with Jenkins?

ofuhrer:
> @mcgibbon @ofuhrer What's up with Jenkins?

Not sure, problem occurs both on GCP as well as Daint when installing `requirements.txt` as far as I can see. @mcgibbon ?
image

ofuhrer:
launch jenkins

nbren12:
It looks like `wheel` needs to be installed. Sometimes older versions of the system python/setuptools don't come with it.

PR #100, opened 2020-09-07
Remove host/shorthost check

ofuhrer:
Remove a check which makes sure the the hostname is contained in host.
Reason: hostname is set by GCP to be `ubuntu18.04-...`.


nbren12 (approved):
LGTM. Could you give a bit more context in the commit message? Are we running jenkins on GCP now?

PR #101, opened 2020-09-10
added diagnostic table for baroclinic test

oelbert:
Just adding a diagnostic table for the baroclinic wave test case


spencerkclark (approved):
This looks good to me.  One minor suggestion might be to include "baroclinic" in the filename of the diag table.

oelbert:
> This looks good to me. One minor suggestion might be to include "baroclinic" in the filename of the diag table.

Good point. Just committed that

PR #103, opened 2020-09-11
Remove network dependency of test suite

mcgibbon:
This PR modifies the test suite to run fully offline. It makes two changes:
- while the tests are running, get_fs is modified so it returns a MemoryFileSystem which accepts paths with a gs protocol
- run_docker tests which actually run the model are removed. This feature is deprecated, and also sufficiently tested by unit tests on run command generation

This was done for a few reasons:
- it is generally good practice to have your tests not depend on network connectivity, to avoid intermittent failures
- we are making our google cloud buckets private, and can no longer depend on the data which was being accessed
- we would like to avoid egress fees when running tests on Piz Daint

### DIFF
--- a/tests/c12_config.yml
+++ b/tests/c12_config.yml
@@ -1,7 +1,7 @@
 data_table: default
 diag_table: default
 experiment_name: default
-forcing: "gs://vcm-fv3config/data/base_forcing/v1.1/"
+forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
 orographic_forcing: gs://vcm-fv3config/data/orographic_data/v1.0
 initial_conditions: gs://vcm-fv3config/data/initial_conditions/gfs_c12_example/v1.0 
 namelist:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,77 @@
 import pytest
 import os
+import io
 import yaml
+import fsspec
+from fsspec.implementations.memory import MemoryFileSystem
+import fv3config.filesystem
+import fv3config.data
 
 TEST_DIR = os.path.dirname(os.path.realpath(__file__))
+
+MOCK_FS_FILENAMES = [
+    "vcm-fv3config/data/initial_conditions/gfs_c12_example/v1.0/initial_conditions_file",
+    "vcm-fv3config/data/base_forcing/v1.1/forcing_file",
+    "vcm-fv3config/data/base_forcing/v1.1/grb/grb_forcing_file",
+    "vcm-fv3config/data/orographic_data/v1.0/C12/orographic_file",
+]
+DIAG_TABLE_PATH = "gs://vcm-fv3config/config/diag_table/default/v1.0/diag_table"
+
+
+def populate_mock_filesystem(fs):
+    for filename in MOCK_FS_FILENAMES:
+        fs.mkdir(os.path.dirname(filename))
+        with fs.open(filename, mode='wb') as f:
+            f.write(b"mock_data")
+    with open(os.path.join(fv3config.data.DATA_DIR, "diag_table/diag_table_default"), "r") as f_in:
+        with fs.open(DIAG_TABLE_PATH, 'w') as f_out:
+            f_out.write(f_in.read())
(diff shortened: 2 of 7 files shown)


tests/c12_config.yml
```diff
@@ -1,7 +1,7 @@
 data_table: default
 diag_table: default
 experiment_name: default
-forcing: "gs://vcm-fv3config/data/base_forcing/v1.1/"
+forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
```
ofuhrer, line 4: Funny, I would've opted to put quotes around these URLs in YAML, mostly because they contain characters with special meaning in YAML such as `:` which can easily lead to unwanted errors. See also [stachoverflow](https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml).

tests/conftest.py
```diff
@@ -1,9 +1,78 @@
...
+        with fs.open(filename, mode="wb") as f:
+            f.write(b"mock_data")
+    with open(
+        os.path.join(fv3config.data.DATA_DIR, "diag_table/diag_table_default"), "r"
+    ) as f_in:
+        with fs.open(DIAG_TABLE_PATH, "w") as f_out:
+            f_out.write(f_in.read())
+    for filename in MOCK_FS_FILENAMES:
+        dirname = os.path.dirname(filename)
+        fs.ls(dirname)
```
ofuhrer, line 32: Seems like this was for testing and can now be removed.

tests/test_forcingdata.py
```diff
@@ -158,7 +125,9 @@ def test_write_default_initial_conditions_directory(self):
         rundir = self.make_run_directory("test_rundir")
         config = DEFAULT_CONFIG.copy()
         asset_list = get_initial_conditions_asset_list(config)
+        print("assets", asset_list)
         write_asset_list(asset_list, rundir)
+        print("asset write dir", os.listdir(rundir))
```
oliverwm1, line 130: leftover print statements?

oliverwm1 (approved):
Nice! Changes LGTM. Agreed the tests of run_docker and run_native are not critical.

tests/conftest.py
```diff
@@ -1,9 +1,78 @@
...
+        path = self._strip_protocol(path)
+        return super().open(path, *args, **kwargs)
+
+
+@pytest.fixture(autouse=True)
+def mock_gs_fs():
+    memory_fs = MockGCSFileSystem()
+    populate_mock_filesystem(memory_fs)
+
+    def mock_get_fs(path: str) -> fsspec.AbstractFileSystem:
```
ofuhrer, line 65: Naming the same on purpose?

tests/test_forcingdata.py
```diff
@@ -158,7 +125,9 @@ def test_write_default_initial_conditions_directory(self):
         rundir = self.make_run_directory("test_rundir")
         config = DEFAULT_CONFIG.copy()
         asset_list = get_initial_conditions_asset_list(config)
+        print("assets", asset_list)
```
ofuhrer, line 128: Debug prints?

tests/test_forcingdata.py
ofuhrer, line 130: Debug prints?

tests/test_fv3run.py
```diff
@@ -124,6 +124,7 @@ def check_run_directory(dirname):
     assert "INPUT" in filenames
 
 
+@pytest.mark.filterwarnings("ignore:could not remove stacksize limit")
```
ofuhrer, line 127: Why does setting `resource.RLIMIT_STACK` not work for this mock test?

ofuhrer (changes requested):
Very elegant solution using an in-memory mock filesystem.
Minor suggestions.

tests/conftest.py
ofuhrer: Sorry was too fast. Maybe more verbose `gs` would be instructive.

ofuhrer (approved):

PR #104, opened 2020-09-15
Add docker file, update enable_restart, add nudging module

oliverwm1:
This PR implements a few changes that are necessary in order to allow segmented fv3gfs runs that include nudging to GFS analysis:

1. Add a Dockerfile to produce a lightweight image with fv3config installed
2. Update enable_restart function to take an optional `initial_conditions` argument and fix a bug in which the function didn't set `force_date_from_namelist` to False (for a restart run, want to read start date from coupler.res file).
3. Add new public API functions `get_nudging_assets` and `update_config_for_nudging`.
4. Add CLI entry points for `enable_restart` and `update_config_for_nudging`.

An example usage of these entrypoint functions can be found in an fv3net branch [here](https://github.com/VulcanClimateModeling/fv3net/blob/c9c6262cac819073097ab8c389745cbbf3368e0b/workflows/argo/run-fv3gfs.yaml#L237-L256).

In order to allow `update_config_for_nudging` to have only the config object as an argument the `get_current_date` function is refactored to not require the path to the `INPUT` directory.

While a convenience Makefile rule is added for image building, the docker image will be automatically built and pushed by Dockerhub from the master branch. See here: https://hub.docker.com/r/vcmml/fv3config


fv3config/config/nudging.py
```diff
@@ -0,0 +1,144 @@
...
+        nudge_filename_pattern: datetime pattern of nudging files. Defaults to
+            '%Y%m%d_%HZ_T85LR.nc'.
+        input_list_filename: name of text file listing all nudging files. Defaults to
+            'nudging_file_list'.
+    """
+    is_nudging_asset = partial(_target_name_matches, pattern=nudge_filename_pattern)
+    return [item for item in assets if not is_nudging_asset(item)]
+
+
+def _target_name_matches(asset, pattern):
```
spencerkclark, line 103: nit: maybe call this `_is_nudging_asset`?  I realize you use a similar name for the partial function above, but I think `_target_name_matches` is a bit vague given its somewhat nudging-file-specific behavior (i.e. checking that the target_name can be parsed for a date with a given time pattern).

spencerkclark (approved):
Looks great!  Just some minor suggestions if you feel inclined.

fv3config/cli.py
```diff
@@ -17,10 +17,63 @@ def _parse_args():
...
-    args = _parse_args()
+    args = _parse_write_run_directory_args()
 
     with fsspec.open(args.config) as f:
         config = yaml.safe_load(f)
 
     fv3config.write_run_directory(config, args.rundir)
+
+
+def enable_restart():
```
mcgibbon, line 54: Please add documentation for these new CLI commands.

setup.py
```diff
@@ -43,6 +36,8 @@
         "console_scripts": [
             "fv3run=fv3config.fv3run.__main__:main",
             "write_run_directory=fv3config.cli:write_run_directory",
+            "enable_restart=fv3config.cli:enable_restart",
+            "update_config_for_nudging=fv3config.cli:update_config_for_nudging",
```
mcgibbon, line 40: Please add these to the docs.

fv3config/config/alter.py
```diff
@@ -4,11 +4,13 @@
...
 
 
-def enable_restart(config):
+def enable_restart(config, initial_conditions=None):
     """Apply namelist settings for initializing from model restart files.
 
     Args:
         config (dict): a configuration dictionary
+        initial_conditions (str): path to desired new initial conditions. Defaults to
+            None, in which case initial condition entry is not modified.
```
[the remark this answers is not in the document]
mcgibbon: I think I agree with Spencer's comment. Either this should be a required argument, or it shouldn't be an argument. The main reason to make it either not an argument or an optional argument is backwards compatibility. Workflows before this have run `config["initial_conditions"] = restart_conditions; enable_restart(config)`.

fv3config/config/nudging.py
```diff
@@ -0,0 +1,144 @@
...
+def _most_recent_hour(current_hour: int, hour_array=NUDGE_HOURS) -> int:
+    """Return latest hour in hour_array that precedes or is concurrent with
+    current_hour"""
+    for hour in hour_array:
+        if hour <= current_hour:
+            first_nudge_hour = hour
+    return first_nudge_hour
+
+
+def _get_nudge_time_list(run_duration, current_date) -> List[datetime]:
```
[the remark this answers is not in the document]
mcgibbon: +1

fv3config/config/nudging.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Args:
+        assets: sequence of assets
+        nudge_filename_pattern: datetime pattern of nudging files. Defaults to
+            '%Y%m%d_%HZ_T85LR.nc'.
+        input_list_filename: name of text file listing all nudging files. Defaults to
+            'nudging_file_list'.
+    """
+    is_nudging_asset = partial(_target_name_matches, pattern=nudge_filename_pattern)
+    return [item for item in assets if not is_nudging_asset(item)]
```
mcgibbon, line 100: Cool!

requirements.txt
```diff
@@ -0,0 +1,7 @@
+f90nml>=1.1.0
+appdirs>=1.4.0
+requests>=2.22.0
+pyyaml>=5.0
+gcsfs>=0.4.0
+fsspec>=0.6.0
```
mcgibbon, line 6: Kind of unrelated, but can you increase these versions to I think 0.6.0 and 0.8.0? Whatever it is in fv3net. The interim versions have caused incompatibility issues when gcsfs is newer and fsspec is older.

tests/test_namelist.py
```diff
@@ -182,8 +184,8 @@ def test_enable_restart_from_default(self):
             restart_namelist_settings, restart_config["namelist"]
         )
 
-    def test_enable_restart_from_empty_fv_core_nml(self):
-        restart_config = enable_restart(config_with_empty_fv_core_nml)
+    def test_enable_restart_from_empty_fv_core_and_coupler_nml(self):
+        restart_config = enable_restart(config_with_empty_fv_core_and_coupler_nml)
```
mcgibbon, line 188: Should this work on an entirely empty namelist? I thought there's no difference from a Fortran standpoint between not defining one of the namelists or defining it to be empty? If we do want to support empty namelists (e.g. where coupler_nml is not present in the `.nml` file at all) please add a test case.

fv3config/cli.py
```diff
@@ -17,10 +17,63 @@ def _parse_args():
...
+    parser = argparse.ArgumentParser("enable_restart")
+    parser.add_argument(
+        "config",
+        help="URI to fv3config yaml file. Supports any path used by fsspec. "
+        "File will be modified in place.",
+    )
+    parser.add_argument(
+        "--initial-conditions",
+        help="Path to restart initial conditions.",
+        default=None,
```
mcgibbon, line 30: By default this routine is going to write a config.yaml which says initial_conditions is None. Please either modify the code so it doesn't modify the initial conditions by default, or make this a required argument.

fv3config/config/alter.py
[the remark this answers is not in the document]
mcgibbon: Even if this function doesn't modify initial_conditions, you can still do it directly in the CLI entrypoint function.

mcgibbon (changes requested):
The test parameterization comments would be good if you can, but are more time intensive so only suggested. Biggest required change is either clarifying NUDGE_HOURS is hard-coded in the docstrings, or allowing it as a keyword argument with a default.

fv3config/config/alter.py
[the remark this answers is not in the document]
oliverwm1: Okay, well I would prefer to require it as an argument and introduce the breaking change. `enable_restart` is not currently used anywhere in the fv3net source tree. Not sure if it's getting used on DSL side.

fv3config/config/nudging.py
oliverwm1: I went back and forth on this, but going to leave it as is, since the pattern is an argument to the function, so it doesn't really know anything about nudging per se.

Makefile
```diff
@@ -90,3 +92,6 @@ dist: clean ## builds source and wheel package
 
 install: clean ## install the package to the active Python's site-packages
 	python3 setup.py install
+
+build_image:
+	docker build -t vcmml/fv3config .
```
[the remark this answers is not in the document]
oliverwm1: Okay, seems like this is going to be a bigger discussion, so I just removed the Makefile target for now.

fv3config/config/nudging.py
```diff
@@ -0,0 +1,144 @@
...
+import math
+import fsspec
+from .._asset_list import get_asset_dict
+from .._exceptions import ConfigError
+from ..filesystem import get_fs
+from .derive import get_current_date, get_run_duration
+
+# this module assumes that analysis files are at 00Z, 06Z, 12Z and 18Z
+SECONDS_IN_HOUR = 60 * 60
+NUDGE_HOURS = [0, 6, 12, 18]  # hours at which analysis data is available
```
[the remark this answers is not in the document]
oliverwm1: I'll make it a keyword arg with a default. The Fortran module can handle different intervals. Although the data we're using currently to nudge is only available every 6h.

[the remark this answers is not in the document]
oliverwm1: I'm actually going to make the keyword arg a single integer for the interval, since the Fortran does require the inputs at regular intervals.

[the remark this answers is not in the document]
oliverwm1: timedelta, not an integer.

fv3config/config/nudging.py
```diff
@@ -0,0 +1,144 @@
...
+    """Return datetime object for the last nudging time preceding or concurrent
+     with start_time"""
+    first_nudge_hour = _most_recent_hour(start_time.hour)
+    return datetime(start_time.year, start_time.month, start_time.day, first_nudge_hour)
+
+
+def _most_recent_hour(current_hour: int, hour_array=NUDGE_HOURS) -> int:
+    """Return latest hour in hour_array that precedes or is concurrent with
+    current_hour"""
+    for hour in hour_array:
```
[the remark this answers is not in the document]
oliverwm1: This no longer applies since the function that calls this one generates the hour_array from an interval so it is guaranteed to be sorted.

tests/test_nudging.py
```diff
@@ -0,0 +1,150 @@
...
+    pattern = "%Y%m%d_%H.nc"
+    assets = fv3config.get_nudging_assets(
+        timedelta(hours=6),
+        [2016, 1, 1, 0, 0, 0],
+        nudge_url,
+        nudge_filename_pattern=pattern,
+    )
+    expected = [
+        fv3config.get_asset_dict(nudge_url, "20160101_00.nc", target_location="INPUT"),
+        fv3config.get_asset_dict(nudge_url, "20160101_06.nc", target_location="INPUT"),
```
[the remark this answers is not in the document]
oliverwm1: But I don't want to test the behavior of `get_asset_dict`. I.e. the function that I'm testing uses `get_asset_dict` and so it seems appropriate to use it in the test as well.

tests/test_nudging.py
```diff
@@ -0,0 +1,150 @@
...
+    with pytest.raises(fv3config.ConfigError):
+        fv3config.get_nudging_assets(
+            timedelta(hours=6),
+            [2016, 1, 1, 0, 0, 0],
+            "gs://bucket/path",
+            copy_method="link",
+        )
+
+
+def test_clear_nudging_assets():
```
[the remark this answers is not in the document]
oliverwm1: Good idea. Handles non-asset items now.

tests/test_namelist.py
oliverwm1: Okay, I allowed the namelist to not contain `fv_core_nml` or `coupler_nml` but the config object still has to have a `namelist` item, which I believe is a requirement in various other places of this package.

fv3config/cli.py
oliverwm1: I actually don't think this was true, but regardless, initial conditions is now a required argument of the `enable_restart` function and a required positional argument of the CLI.

mcgibbon: OK, sorry for the misread and thanks for the update.

fv3config/config/alter.py
```diff
@@ -4,26 +4,29 @@
...
     Returns:
         dict: a configuration dictionary
     """
     if "namelist" not in config:
         raise ConfigError("config dictionary must have a 'namelist' key")
-    if "fv_core_nml" not in config["namelist"]:
-        raise ConfigError("config dictionary must have a 'fv_core_nml' namelist")
     restart_config = copy.deepcopy(config)
-    restart_config["namelist"]["fv_core_nml"]["external_ic"] = False
+    restart_config["namelist"].setdefault("fv_core_nml", {})["external_ic"] = False
```
mcgibbon, line 20: I learned a new dict method today.

fv3config/config/nudging.py
```diff
@@ -0,0 +1,152 @@
...
+def _most_recent_hour(current_hour: int, hour_array: Sequence[int]) -> int:
+    """Return latest hour in hour_array that precedes or is concurrent with
+    current_hour"""
+    for hour in hour_array:
+        if hour <= current_hour:
+            first_nudge_hour = hour
+    return first_nudge_hour
+
+
+def _get_nudge_time_list(
```
mcgibbon, line 32: Nicely done!

mcgibbon (approved):
LGTM. Thanks for the refactors!

PR #107, opened 2020-10-06
Check for coupler.res among all assets

oliverwm1:
Check for the existence of `coupler.res` amongst all assets of given configuration dictionary. This fixes a bug introduced in #104 which required `initial_conditions` to be a path, whereas it can also be an asset or list of assets.

As part of this:
1. move `get_resolution` from `derive.py` to `_datastore.py` (this is the only module in which this function is used)
1. add two tests of `get_current_date` which exercise the two code paths of that function which are not currently tested

### DIFF
--- a/fv3config/_datastore.py
+++ b/fv3config/_datastore.py
@@ -1,7 +1,6 @@
 import os
 
 from .caching import get_internal_cache_dir
-from .config.derive import get_resolution
 from .data import DATA_DIR
 from ._exceptions import ConfigError
 from . import filesystem
@@ -20,6 +19,28 @@ FIELD_TABLE_OPTIONS = {
     "GFDLMP": "field_table_GFDLMP",
     "ZhaoCarr": "field_table_ZhaoCarr",
 }
+
+
+def get_resolution(config):
+    """Get the model resolution based on a configuration dictionary.
+
+    Args:
+        config (dict): a configuration dictionary
+
+    Returns:
+        resolution (str): a model resolution (e.g. 'C48' or 'C96')
+
+    Raises:
+        ConfigError: if the number of processors in x and y on a tile are unequal
+    """
+    npx = config["namelist"]["fv_core_nml"]["npx"]
+    npy = config["namelist"]["fv_core_nml"]["npy"]
+    if npx != npy:
+        raise ConfigError(
+            f"npx and npy in fv_core_nml must be equal, but are {npx} and {npy}"
+        )
+    resolution = f"C{npx-1}"
+    return resolution
 
 
 def get_orographic_forcing_directory(config):
--- a/tests/test_nudging.py
+++ b/tests/test_nudging.py
@@ -152,7 +152,6 @@ def test__is_nudging_asset(item, pattern, expected):
 
 
 def test_update_config_for_nudging(test_config):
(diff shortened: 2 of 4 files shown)


fv3config/_datastore.py
```diff
@@ -22,6 +21,28 @@
 }
 
 
+def get_resolution(config):
```
mcgibbon, line 24: Was there a reason to move this from the public `derive` namespace to this private namespace? The idea of `fv3config.config.derive` (whose members should be imported to the top level instead, probably my bad) is to provide routines which derive data from a configuration dictionary (excluding data paths). If there isn't a need for it to be moved, I'd suggest keeping it in the module it was in.

tests/test_nudging.py
```diff
@@ -146,22 +151,20 @@ def test__is_nudging_asset(item, pattern, expected):
     assert nudging._is_nudging_asset(item, pattern) == expected
 
 
-def test_update_config_for_nudging():
+def test_update_config_for_nudging(test_config):
```
mcgibbon, line 154: Good change!

tests/test_tables.py
```diff
@@ -76,19 +77,19 @@ def make_run_directory(self, directory_name):
         return full_path
 
     def test_default_data_table_filename(self):
-        config = DEFAULT_CONFIG.copy()
+        config = copy.deepcopy(DEFAULT_CONFIG)
```
mcgibbon, line 80: Good

tests/test_tables.py
```diff
@@ -161,13 +162,35 @@ def test_get_current_date_from_bad_coupler_res(self):
...
-    def test_get_current_date_from_config(self):
-        config = DEFAULT_CONFIG.copy()
+    def test_get_current_date_from_config_force_date_true(self):
+        config = copy.deepcopy(DEFAULT_CONFIG)
         config["namelist"]["coupler_nml"]["force_date_from_namelist"] = True
         config["namelist"]["coupler_nml"]["current_date"] = valid_current_date
         current_date = get_current_date(config)
         self.assertEqual(current_date, valid_current_date)
 
+    def test_get_current_date_from_config_force_date_false(self):
```
mcgibbon, line 172: If it's not too much trouble, could you add a case where `initial_conditions` is an empty list `[]`? This is useful for idealized cases which need no initial conditions data, and is an interesting edge case.

mcgibbon (approved):
Thanks for the quick turnaround! Two suggestions.

fv3config/_datastore.py
oliverwm1: It was necessary to prevent a circular import issue. Currently, `get_resolution` is not imported by any `__init__`'s and is only used in `_datastore.py` so it seemed like that was a more appropriate place.

mcgibbon: OK, we can consider moving it later then.

tests/test_tables.py
oliverwm1: Case added.

fv3config/_datastore.py
oliverwm1: Sounds good.

oliverwm1:
Resolves #108

PR #111, opened 2020-10-27
Fix getting coupler_res filename when using directory for initial conditions

oliverwm1:
Fixes bug (and adds test) for getting coupler res filename when using a directory for initial conditions. Bug happened because if statement checked if `target_location` was equal to `INPUT` but `asset_list_from_path` returned `INPUT/`. Fixes bug by using joined path in `_get_coupler_res_filename`.


frodre (approved):
Thanks for the quick fix!

PR #112, opened 2020-11-05
Explicitly install wheel before requirements to remove wheel error messages

ofuhrer:
While not failling, tests are cluttered with error messages along the lines `error: invalid command 'bdist_wheel'`. This fixes this by pre-pip-installing wheel.


mcgibbon (approved):
LGTM, thanks for the fix!

PR #113, opened 2020-11-06
Fix date retrieval from coupler_res to use get_fs instead of fsspec.open

frodre:
Use the fv3config `get_fs` to instantiate the fileststem used to grab the coupler.res file when getting the date date instead of `fsspec.open`.  This ensures that anything in the public `fv3config` bucket will have the `requester_pays=True` in the `GCSFSFilesystem` creation.


AnnaKwa (approved):
LGTM

PR #118, opened 2021-01-22
Validate md5 checksum of gcs downloads

### LINKED ISSUE #117: Verify checksums for downloaded files
Occasionally the downloads of certain will be corrupted. Unfortunately, future invocations of write run directory will not fix this problem, since the corrupted data will be stored in the cache. We should verify the correctness of downloaded files before storing them in the cache. I believe that every object in GCS already has a CRC that can be verified.

nbren12:
I've been getting corrupted file errors occasionally. Needed to bump gcs version to 0.7.0 to get the `gcsfs.utils.ChecksumError`.

Resolves #117

### DIFF
--- a/fv3config/filesystem.py
+++ b/fv3config/filesystem.py
@@ -9,7 +9,7 @@ from concurrent.futures import ThreadPoolExecutor, Executor
 try:
     import gcsfs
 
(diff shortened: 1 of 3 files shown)


oliverwm1 (approved):
LGTM.

PR #122, opened 2021-02-02
Add logging to writing of run directory

### LINKED ISSUE #115: Add logging to write_run_directory
Writing a run directory can take a while (often 5+ minutes for some of my cases) when there are many large assets. It would be useful if fv3config provided some logging statements that showed the progress of `write_run_directory` and signaled successful completion of that step.

oliverwm1:
This will aid in debugging if the `write_run_directory` command fails. It is also useful for monitoring, especially when many assets are being copied and `write_run_directory` takes minutes to complete.

Resolves #115 

Edit: also fixes an unrelated warning about importing from collections instead of collections.abc

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,10 @@
 History
 =======
 
+Latest
+------
+
+- Add logging to write_run_directory command
 
 v0.5.1
 ------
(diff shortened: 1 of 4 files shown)


rheacangeo (approved):
Excellent addition! In some situations I end up stuck in an infinite loop where write_run_directory copies the contents of pwd into the rundir/INPUTS dir and I haven't figured out what causes it. Logging might help and surely will help other situations as well!

PR #125, opened 2021-02-16
DiagTable class

oliverwm1:
It is useful to have a richer representation of the `diag_table` item in fv3config configuration dictionaries.

This PR adds a `DiagTable` class and associated `DiagFileConfig` and `DiagFieldConfig` dataclasses. The `DiagTable` class has `__repr__` and `from_str` methods which allow going between the Fortran-expected string representation of the `diag_table` and the python representation. It also includes `asdict` and `from_dict` methods, which are useful for serializing/deserializing the `diag_table` item to/from YAML.

[dacite](https://github.com/konradhalas/dacite) is added as a requirement, as it provides useful functionality for going from a dict to a DataClass with type checking.

This PR does not introduce any breaking changes. If a user provides a path for the `diag_table`, that file will just be copied to the rundir as is currently done.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -228,10 +228,9 @@ The same ``DiagTable`` can also be initialized programmatically as follows:
 String representations of the ``diag_table`` (i.e. those expected by the Fortran model) can be parsed
 with the :py:meth:`fv3config.DiagTable.from_str` method.
 
-If serializing an ``fv3config`` configuration object to yaml, it is recommended to use the
-:py:meth:`fv3config.DiagTable.asdict` method before dumping the configuration dictionary. This
-method is called by :py:meth:`fv3config.config_to_yaml` if the ``diag_table`` item of the
-configuration dictionary is a ``DiagTable`` instance.
+If serializing an ``fv3config`` configuration object to yaml it is recommended to use
+:py:meth:`fv3config.config_to_yaml`. This method will convert any ``DiagTable`` instances to
+dicts (using :py:meth:`fv3config.DiagTable.asdict`), which can be safely serialized.
 
 
 Running the model with fv3run
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -158,7 +158,7 @@ for the string representation of the ``diag_table``
 package defines a python representation of this object, :py:class:`~fv3config.DiagTable`, which can
 be used to explicitly represent the ``diag_table`` within an fv3config configuration dictionary.
 
-The ``DiagTable`` object can be initialized from a dict (serialized as YAML) as follows. Suppose
+The ``DiagTable`` object can be initialized from a dict (here serialized as YAML) as follows. Suppose
 the following is saved within ``sample_diag_table.yaml``:
 
 .. code-block:: yaml
@@ -177,7 +177,7 @@ the following is saved within ``sample_diag_table.yaml``:
         module_name: gfs_phys
         output_name: upward_longwave_radiative_flux_at_toa
 
-Then:
+Then a ``DiagTable`` object can be initialized as:
 
 .. code-block:: python
 
@@ -202,35 +202,36 @@ The same ``DiagTable`` can also be initialized programmatically as follows:
     >>> import fv3config
     >>> import datetime
     >>> diag_table = fv3config.DiagTable(
-        name="example_diag_table",
-        base_time=datetime.datetime(2000, 1, 1)
-        file_configs=[
-            fv3config.DiagFileConfig(
-                name="physics_diagnostics",
-                frequency=1,
-                frequency_units="hours",
-                field_configs=[
-                    fv3config.DiagFieldConfig(
-                        "gfs_phys",
-                        "totprcb_ave"
-                        "surface_precipitation_rate"
-                    ),
-                    fv3config.DiagFieldConfig(
-                        "gfs_phys",
-                        "ULWRFtoa"
-                        "upward_longwave_radiative_flux_at_toa"
-                    ),
-                ]
-            )
-        ]
-    )
+            name="example_diag_table",
+            base_time=datetime.datetime(2000, 1, 1)
+            file_configs=[
+                fv3config.DiagFileConfig(
+                    name="physics_diagnostics",
+                    frequency=1,
+                    frequency_units="hours",
+                    field_configs=[
+                        fv3config.DiagFieldConfig(
+                            "gfs_phys",
+                            "totprcb_ave"
+                            "surface_precipitation_rate"
+                        ),
+                        fv3config.DiagFieldConfig(
+                            "gfs_phys",
+                            "ULWRFtoa"
+                            "upward_longwave_radiative_flux_at_toa"
+                        ),
+                    ]
+                )
+            ]
+        )
 
-String representations of the `diag_table` (i.e. those expected by the Fortran model) can be parsed
+String representations of the ``diag_table`` (i.e. those expected by the Fortran model) can be parsed
 with the :py:meth:`fv3config.DiagTable.from_str` method.
 
-If serializing an `fv3config` configuration object to yaml, it is recommended to use the
-:py:meth:`fv3config.DiagTable.asdict` method before dumping the configuration dictionary. This is called by
-by :py:meth:`fv3config.config_to_yaml`.
+If serializing an ``fv3config`` configuration object to yaml, it is recommended to use the
+:py:meth:`fv3config.DiagTable.asdict` method before dumping the configuration dictionary. This
+method is called by :py:meth:`fv3config.config_to_yaml` if the ``diag_table`` item of the
+configuration dictionary is a ``DiagTable`` instance.
 
 
 Running the model with fv3run
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -75,17 +75,17 @@ Configuration
 
 The ``config`` dictionary must have at least the following items:
 
-==================== ======== ============================================
-Key                  Type     Description
-==================== ======== ============================================
-namelist             dict     Model namelist
-experiment_name      str      Name of experiment to use in output
-diag_table           str      location of diag_table file, or one of ("default", "grid_spec", "no_output")
-data_table           str      location of data_table file, or "default"
-initial_conditions   str      location of directory containing initial conditions data
-forcing              str      location of directory containing forcing data
-orographic_forcing   str      location of directory containing orographic data
-==================== ======== ============================================
+==================== ======================================== ============================================
+Key                  Type                                     Description
+==================== ======================================== ============================================
+namelist             dict                                     Model namelist
+experiment_name      str                                      Name of experiment to use in output
+diag_table           str or :py:class:`~fv3config.DiagTable`  location of diag_table file, or one of ("default", "grid_spec", "no_output"), or DiagTable object
+data_table           str                                      location of data_table file, or "default"
+initial_conditions   str                                      location of directory containing initial conditions data
+forcing              str                                      location of directory containing forcing data
+orographic_forcing   str                                      location of directory containing orographic data
+==================== ======================================== ============================================
 
 Paths to files or directories on the local
 filesystem must be given as absolute paths. If paths are given that begin with ``gs://`` then ``fv3config`` will
@@ -95,6 +95,10 @@ must be installed and authorized to download from the specified bucket.
 The ``namelist`` item is special in that it is explicitly stored in the ``config`` dictionary. For the
 fv3gfs model, individual namelists are specified for various components of the model. As an example, the
 vertical resolution can be accessed via ``config['namelist']['fv_core_nml']['npz']``.
+
+The ``diag_table`` can be either be a tag or path to a file, or it can explicitly represent
+the desired output diagnostics with a :py:class:`~fv3config.DiagTable` object. See a more complete
+description of this object below.
 
 By default, fv3config attempts to automatically select the ``field_table`` file
 to use for the model based on the selected microphysics scheme in the
@@ -144,6 +148,60 @@ One can use a directory to specify the initial conditions or forcing files and r
 subset of the files within the that directory with the optional ``config['patch_files']`` item.
 All assets defined in ``config['patch_files']`` will overwrite any files specified in the
 initial conditions or forcing if they have the same target location and name.
+
+DiagTable configuration
+-----------------------
+
+The ``diag_table`` specifies the diagnostics to be output by the Fortran model. See documentation
+for the string representation of the ``diag_table``
+`here <https://mom6.readthedocs.io/en/latest/api/generated/pages/Diagnostics.html>`_. The fv3config
+package defines a python representation of this object, :py:class:`~fv3config.DiagTable`, which can
+be used to explicitly represent the ``diag_table`` within an fv3config configuration dictionary.
+
+The ``DiagTable`` object can be initialized from a dict as follows:
+
+.. code-block:: python
+
+    >>> import fv3config
+    >>> from datetime import datetime
+    >>> diag_table_dict = {
+        "name": "example_diag_table",
+        "base_time": datetime(2000, 1, 1),
+        "files": [
+            {
+                "name": "physics_diagnostics",
+                "frequency": 1,
+                "frequency_units": "hours",
+                "fields": [
+                    {
+                        "module_name": "gfs_phys",
+                        "field_name": "totprcpb_ave",
+                        "output_name": "surface_precipitation_rate",
+                    },
+                    {
+                        "module_name": "gfs_phys",
+                        "field_name": "ULWRFtoa",
+                        "output_name": "upward_longwave_radiative_flux_at_toa",
+                    }
+                ]
+            }
+        ]
+    }
+    >>> diag_table = fv3config.DiagTable.from_dict(diag_table_dict)
+    >>> print(diag_table)  # will output diag_table format expected by Fortran model
+    example_diag_table
+    2000 1 1 0 0 0
+
+    "physics_diagnostics", 1, "hours", 1, "hours", "time"
+
+    "gfs_phys", "totprcpb_ave", "surface_precipitation_rate", "physics_diagnostics", "all", "none", "none", 2
+    "gfs_phys", "ULWRFtoa", "upward_longwave_radiative_flux_at_toa", "physics_diagnostics", "all", "none", "none", 2
+
+String representations of the `diag_table` can be parsed with the :py:meth:`fv3config.DiagTable.from_str` method.
+
+If serializing an `fv3config` configuration object to yaml, it is recommended to use the
+:py:meth:`fv3config.DiagTable.asdict` method before dumping the configuration dictionary. This is implemented
+by :py:meth:`fv3config.config_to_yaml`.
 
 
 Running the model with fv3run
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -158,35 +158,33 @@ for the string representation of the ``diag_table``
 package defines a python representation of this object, :py:class:`~fv3config.DiagTable`, which can
 be used to explicitly represent the ``diag_table`` within an fv3config configuration dictionary.
 
-The ``DiagTable`` object can be initialized from a dict as follows:
+The ``DiagTable`` object can be initialized from a dict (serialized as YAML) as follows. Suppose
+the following is saved within ``sample_diag_table.yaml``:
+
+.. code-block:: yaml
+
+    name: example_diag_table
+    base_time: 2000-01-01 00:00:00
+    file_configs:
+    - name: physics_diagnostics
+      frequency: 1
+      frequency_units: hours
+      field_configs:
+      - field_name: totprcpb_ave
+        module_name: gfs_phys
+        output_name: surface_precipitation_rate
+      - field_name: ULWRFtoa
+        module_name: gfs_phys
+        output_name: upward_longwave_radiative_flux_at_toa
+
+Then:
 
 .. code-block:: python
 
     >>> import fv3config
-    >>> from datetime import datetime
-    >>> diag_table_dict = {
-        "name": "example_diag_table",
-        "base_time": datetime(2000, 1, 1),
-        "files": [
-            {
-                "name": "physics_diagnostics",
-                "frequency": 1,
-                "frequency_units": "hours",
-                "fields": [
-                    {
-                        "module_name": "gfs_phys",
-                        "field_name": "totprcpb_ave",
-                        "output_name": "surface_precipitation_rate",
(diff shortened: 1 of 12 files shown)


fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+@dataclasses.dataclass
+class DiagTableField:
+    """Object representing a field of a diagnostics file."""
+
+    module_name: str
+    field_name: str
+    output_name: str
+    reduction_method: Optional[str] = "none"
```
oliverwm1, line 21: with Python 3.8, could use `Literal["none", "average", "min", "max"]` as type hint, which would make for awesome self-documentation and runtime validation.

docs/usage.rst
```diff
@@ -145,6 +149,60 @@ subset of the files within the that directory with the optional ``config['patch_
...
 initial conditions or forcing if they have the same target location and name.
 
+DiagTable configuration
+-----------------------
+
+The ``diag_table`` specifies the diagnostics to be output by the Fortran model. See documentation
+for the string representation of the ``diag_table``
+`here <https://mom6.readthedocs.io/en/latest/api/generated/pages/Diagnostics.html>`_. The fv3config
+package defines a python representation of this object, :py:class:`~fv3config.DiagTable`, which can
+be used to explicitly represent the ``diag_table`` within an fv3config configuration dictionary.
```
mcgibbon, line 159: Maybe out of scope for this PR, but we should define some kind of internal class (not a dict) which explicitly types the fully-processed configuration (which we currently use a dict for). As it is, the term "configuration dictionary" is overloaded, since it can mean the output of yaml.load (which is more loosely typed) or the processed version of this.

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+    time_sampling: Optional[str] = "all"
+    regional_section: Optional[str] = "none"
+    packing: Optional[int] = 2
+
+
+@dataclasses.dataclass
+class DiagTableFile:
+    """Object representing a diagnostics file."""
```
mcgibbon, line 29: "Object representing a diagnostics file" and the name DiagTableFile originally had me think this represented a file containing diagnostic data. Can this be renamed to fully convey what it is? I'd suggest DiagFileSpecification, DiagFileOutputSpec, or something else. If you use Spec or Specification I would suggest making a similar change/using the same word for the DiagTableField class.

It was hard for me to understand some of the documentation initially because of this confusion.

The Table part can be dropped, since on its own this class has nothing to do with a table.

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+
+    name: str
+    frequency: int
+    frequency_units: str
+    fields: Sequence[DiagTableField]
+    file_format: Optional[int] = 1
```
mcgibbon, line 35: Can/should this be implemented as an [Enum class](https://docs.python.org/3/library/enum.html)? I can't tell what this means as-is.

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+    time_axis_units: Optional[str] = "hours"
+    time_axis_name: Optional[str] = "time"
```
mcgibbon, line 37: Is this referring to a netCDF coordinate? I'm supportive of keeping the same naming as in the diag_table documentation, but you may want to add a docstring if you know the answer to this.

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+    packing: Optional[int] = 2
+
+
+@dataclasses.dataclass
+class DiagTableFile:
+    """Object representing a diagnostics file."""
+
+    name: str
+    frequency: int
+    frequency_units: str
```
mcgibbon, line 33: Because time frequency can only be one of "years", "months", "days", "minutes", "hours", or "seconds", I would suggest using an [enum class](https://docs.python.org/3/library/enum.html) for this. This could also be used for the time_axis, if they have the same possibilities. I believe this is [supported by dacite](https://github.com/konradhalas/dacite/pull/65).

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+            name: label used as attribute in output diagnostic files. Cannot contain
+                spaces.
+            base_time: time to be used as reference for time coordinate units.
+            files: sequence of DiagTableFile's defining the diagnostics to be output.
+        """
+        if " " in name:
+            raise ConfigError(f"Name for diag_table cannot have spaces. Got '{name}'.")
+        self.name = name
+        self.base_time = base_time
+        self.files = files
```
mcgibbon, line 62: Can this be renamed to reflect that it is a specification and not a file itself? "files" is ambiguous, and my first reading was that it is an object representing an existing file.

tests/test_diag_table.py
```diff
@@ -0,0 +1,111 @@
+from datetime import datetime
+import pytest
+
+from fv3config.config.diag_table import DiagTable, DiagTableField, DiagTableFile
+from fv3config._exceptions import ConfigError
+
```
mcgibbon, line 6: Thanks for the pretty extensive tests! One case I see missing is ensuring that the DiagTable resolves down to an expected string. Currently, a bug where the DiagTable outputs an incorrectly formatted string will not cause the tests to fail, so long as the DiagTable can also load itself from that string. Since ensuring the DiagTable is exactly as required by the Fortran model is an important use case, please add a couple test cases for this.

requirements_dev.txt
```diff
@@ -20,3 +20,4 @@ google-cloud-storage
 fsspec==0.8.0
 kubernetes==10.0.1
 black==19.10b0
+dacite==1.6.0
```
mcgibbon, line 23: We really should change this repo over to the pip-compile method of pinning packages, if you're interested in doing that at any point in the future (not for this PR).

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+                    frequency_units=file_tokens[2],
+                    fields=fields,
+                    file_format=file_tokens[3],
+                    time_axis_units=file_tokens[4],
+                    time_axis_name=file_tokens[5],
+                )
+            )
+        name = lines[0]
+        base_time = cls._str_to_time(lines[1])
+        return cls(name, base_time, files)
```
mcgibbon, line 212: This function is very large and has some clear taxonomies in what it's doing. Could you separate it into some well-named functions? After reading a while I think I see that the function splits the lines into some representation of file and field lines, and then for each of the files constructs a file object, the list of which can then be used to construct the class. If you put this in a small number of lines, I'd be more confident that's what it's doing (and I'd reach that conclusion more quickly).

docs/usage.rst
```diff
@@ -145,6 +149,60 @@ subset of the files within the that directory with the optional ``config['patch_
 All assets defined in ``config['patch_files']`` will overwrite any files specified in the
 initial conditions or forcing if they have the same target location and name.
 
+DiagTable configuration
+-----------------------
+
+The ``diag_table`` specifies the diagnostics to be output by the Fortran model. See documentation
+for the string representation of the ``diag_table``
+`here <https://mom6.readthedocs.io/en/latest/api/generated/pages/Diagnostics.html>`_. The fv3config
```
mcgibbon, line 157: Great job referring to/making use of existing documentation!

docs/usage.rst
```diff
@@ -145,6 +149,60 @@ subset of the files within the that directory with the optional ``config['patch_
...
+    }
+    >>> diag_table = fv3config.DiagTable.from_dict(diag_table_dict)
+    >>> print(diag_table)  # will output diag_table format expected by Fortran model
+    example_diag_table
+    2000 1 1 0 0 0
+
+    "physics_diagnostics", 1, "hours", 1, "hours", "time"
+
+    "gfs_phys", "totprcpb_ave", "surface_precipitation_rate", "physics_diagnostics", "all", "none", "none", 2
+    "gfs_phys", "ULWRFtoa", "upward_longwave_radiative_flux_at_toa", "physics_diagnostics", "all", "none", "none", 2
```
mcgibbon, line 198: There are two main ways we should encourage using this object: loading a yaml file from disk containing this info and turning it into this class using `from_dict(yaml.load())`, and programmatically constructing this class using its object types and `__init__`.

The second case is safer than programmatically making a dict, because with a dict you only catch errors at the final step of putting the class together, and it's not obvious where you'll get errors, whereas making e.g. file specification objects, you get an error exactly at the file specification you formatted incorrectly.

Could you refactor this into two examples, one showing the dict here as a yaml file and code that loads that yaml file from disk to make this class, and a second part where you show how to initialize the file specification and how the class can be initialized from a sequence of file specifications?

docs/usage.rst
```diff
@@ -145,6 +149,60 @@ subset of the files within the that directory with the optional ``config['patch_
...
+
+String representations of the `diag_table` can be parsed with the :py:meth:`fv3config.DiagTable.from_str` method.
+
+If serializing an `fv3config` configuration object to yaml, it is recommended to use the
+:py:meth:`fv3config.DiagTable.asdict` method before dumping the configuration dictionary. This is implemented
+by :py:meth:`fv3config.config_to_yaml`.
```
mcgibbon, line 204: When you say `implemented by`, do you mean `called by`?

Can this be shown as an example? It would be pretty short, and makes it clear whether it's e.g. `yaml.dump(diag_table.asdict())` or `yaml.dump(DiagTable.asdict(diag_table))`.

fv3config/config/diag_table.py
mcgibbon: This could also be an Enum class.

docs/usage.rst
mcgibbon: The latter is especially important if you use Enum classes, since while they're well-documented and pretty easy to use, most scientists won't have run into them before.

mcgibbon (changes requested):
Thanks for the class-ification of this! My requests fall under:
- emphasising two more likely use cases in the docs of how this class should be constructed
- using Enum classes throughout for choice selection attributes (these seem to be supported by dacite, which is awesome, and add a lot of safety)
- Renaming some ambiguous names by adding some variant of "specification" or "spec"
- splitting up a large method
- adding test cases which ensure the output diag table string is correctly formatted

I think these are reasonable, but let me know if you think any of these can't be done or are too onerous.

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+
+        lines.append("")
+        for file_ in self.files:
+            for field in file_.fields:
+                lines.append(self._field_repr(field, file_.name))
+            lines.append("")
+
+        return "\n".join(lines)
+
+    def asdict(self):
```
mcgibbon, line 82: I know this is mirroring dataclasses, but since we also have `from_dict` with pothole case can this use pothole case for consistency with itself/the rest of our user code?

mcgibbon:
I also should say, really good taxonomic breakdown of the diag table file into file and field specifications. It's just right.

nbren12:
We cannot use python3.8 btw. All the prognostic run images and the fv3gfs-fortran use python 3.6.

fv3config/config/diag_table.py
nbren12: We can't upgrade to 3.8. All the fv3 docker images we use are based on ubuntu:18.04 which has python 3.6.

fv3config/config/diag_table.py
nbren12: I got confused a bit about this naming as well. Over in fv3net, there is a distinction between `*Config` objects which the users configure directly, and the underlying objects which these configurations are constructed into.

nbren12: Instead of `Spec` I would opt for the term `Config` for consistency with the prognostic run runtime configurations.

oliverwm1: Good call, will add `Spec`.

I used `Table` because these classes are tightly coupled to the Fortran `diag_table`. E.g. I don't think we'd use these exact classes if we weren't trying to represent the `diag_table`.

fv3config/config/diag_table.py
nbren12, line 21: Do your really mean `Optional` here? In terms of typing this is roughly equivalent to `Union[str, None]`, which wouldn't make sense unless you intend for the user to provide something like `reduction_method: null` in a yaml file.

If you are just saying this option as a default, you don't need `Optional`.

```suggestion
    reduction_method: str = "none"
```

fv3config/config/diag_table.py
oliverwm1: We might add a diagnostics manager for python diagnostics to fv3config, and it would be useful to have the `Table` distinction at that point.

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+
+@dataclasses.dataclass
+class DiagTableField:
+    """Object representing a field of a diagnostics file."""
+
+    module_name: str
+    field_name: str
+    output_name: str
+    reduction_method: Optional[str] = "none"
+    time_sampling: Optional[str] = "all"
```
nbren12, line 22:
```suggestion
    time_sampling: str = "all"
```

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+    regional_section: Optional[str] = "none"
```
nbren12, line 23:
```suggestion
    regional_section: str = "none"
```

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+    packing: Optional[int] = 2
```
nbren12, line 24:
```suggestion
    packing: int = 2
```

docs/usage.rst
oliverwm1: Yeah agreed. Typing the whole namelist in detail might be some work (and potentially prone to bugs) but it would make configuration easier.

oliverwm1: I do think out of scope for this PR.

fv3config/config/diag_table.py
oliverwm1: Sure I can add a docstring/comment.

tests/test_diag_table.py
oliverwm1: I figured this case was covered by explicitly testing `from_str` and testing the `__repr__` --> `from_str` round trip. But I can add an explicit test for it.

fv3config/config/diag_table.py
oliverwm1: Ah nice, thanks.

fv3config/config/diag_table.py
nbren12: Or even better use a datetime.timedelta object and define a custom parser for frequencies:

```
>>> import dacite
... import dataclasses
... import datetime
...
...
... def parse_str(duration):
...     return datetime.timedelta(hours=3)
...
...
... @dataclasses.dataclass
... class A:
...     freq: datetime.timedelta
...
... a = dacite.from_dict(A, {"freq": "3h"}, config=dacite.Config(type_hooks={datetime.timedelta: parse_str}))
>>> a
A(freq=datetime.timedelta(seconds=10800))
```

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,212 @@
...
+        """Initialize DiagTable class from Fortran string representation."""
+        lines = diag_table.split("\n")
+        lines = [line for line in lines if line and line[0] != "#"]
+        parsed_lines = list(map(cls._parse_line, lines[2:]))
+
+        file_lines = []
+        field_lines = {}
+
+        for tokens in parsed_lines:
+            if len(tokens) == 6:
```
nbren12, line 164: Magic number 6? Could you give this a variable name?

fv3config/config/diag_table.py
oliverwm1: Certain frequency int values such as 0 or -1 have special meanings for the diag_table, which is why I kept the `frequency` and `frequency_units` attributes separate.

nbren12: Couldn't you identify those 0 with `0hr` and -1 with any negative value of `timedelta`?

fv3config/config/diag_table.py
nbren12: Looks like we cross streams. I'd still vote for "Config" , since `Spec` is a synonym and `config` is the existing convention (e.g this package is named`fv3config` not `fv3spec`).

oliverwm1: I agree.

fv3config/config/diag_table.py
oliverwm1: We could... but to me it just introduces more levels of documentation required, e.g. specifying how we expect the timedelta to be serialized ("3h") whereas right now it's clear.

Also the Fortran model has a concept of a "month" time frequency/interval which is not covered by python datetime.timedelta.

oliverwm1: Put another way, the docs available for the diag_table are actually quite good (e.g. https://mom6.readthedocs.io/en/dev-gfdl/api/generated/pages/Diagnostics.html) and so I'm trying to stick to the established spec as much as possible, but with a more hierarchal structure and with sensible defaults. That way we don't have to redocument or learn a new system.

fv3config/config/diag_table.py
mcgibbon: Let's avoid upgrading for now but put this in the box of minor reasons to upgrade our ubuntu version.

fv3config/config/diag_table.py
mcgibbon: I'm good with `Config`.

fv3config/config/diag_table.py
mcgibbon: I love replacing Fortran "time amount plus units" with timedeltas. If we're not refactoring the "special meanings" into a different attribute representation (like boolean flags) and want to stick to the existing docs, I'm supportive of keeping this as-is.

If we want a timedelta, it may be worthwhile adding a property method which does the unit string parsing and provides a timedelta (but raises an exception for months).

tests/test_diag_table.py
mcgibbon: I saw that's what you were going for. Unfortunately it's very possible to write code that is lax in how it interprets the input and accepts a wide variety of formatting, and also write a to-string method which was too lax in its interpretation of the standard and has differences in formatting.

docs/usage.rst
oliverwm1: Okay, I added the two examples you suggested.

fv3config/config/diag_table.py
oliverwm1: Done.

fv3config/config/diag_table.py
oliverwm1: I would prefer to be consistent with the standard library here, although I agree it's a bit unsatisfying.

oliverwm1:
Okay I think I've addressed all your comments @mcgibbon 

> emphasising two more likely use cases in the docs of how this class should be constructed

done

> using Enum classes throughout for choice selection attributes (these seem to be supported by dacite, which is awesome, and add a lot of safety)

done for `int` types (Packing, FileFormat) but not others, since the string descriptions are already clear. The possible options are described in docstrings, and if/when we move to Python 3.8, we could type hint with `Literal`.

> Renaming some ambiguous names by adding some variant of "specification" or "spec"

added `Config`

> splitting up a large method

done

> adding test cases which ensure the output diag table string is correctly formatted

done

The other significant changes are related to `config_to_yaml` and `config_from_yaml`. I made the former a public function, and added tests for both, and tested that the serialization/deserialization of the `diag_table` config item works as expected.

docs/usage.rst
```diff
@@ -145,6 +149,90 @@ subset of the files within the that directory with the optional ``config['patch_
...
+                    ]
+                )
+            ]
+        )
+
+String representations of the ``diag_table`` (i.e. those expected by the Fortran model) can be parsed
+with the :py:meth:`fv3config.DiagTable.from_str` method.
+
+If serializing an ``fv3config`` configuration object to yaml, it is recommended to use the
+:py:meth:`fv3config.DiagTable.asdict` method before dumping the configuration dictionary. This
```
mcgibbon, line 232: Why is this recommended, and what are the dangers if you do not do this? What exactly does "use the asdict method before dumping" mean? Do you mean to replace the diagtable with a dict representation using asdict?

You may want to recommend foremost that the reader uses `config_to_yaml`, and that if they need to serialize manually then they can use asdict to get the primitive representation.

No change required for merge but if you think you can quickly make it clearer please do.

mcgibbon (approved):
Looking good! Please either remove or test `__eq__`, and I had one comment/question/suggestion for the documentation. Neither of these should need me to re-review, so I've approved.

The API looks really nice, I'm excited to have more of these strictly-typed configuration classes within our model code (e.g. prognostic run script, or the DSL dycore).

fv3config/config/diag_table.py
```diff
@@ -0,0 +1,278 @@
...
+        lines.append("")
+        for file_ in self.file_configs:
+            for field in file_.field_configs:
+                lines.append(self._field_repr(field, file_.name))
+            lines.append("")
+
+        return "\n".join(lines)
+
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self.__repr__() == other.__repr__()
```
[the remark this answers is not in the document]
oliverwm1: Okay sure, it made some of the tests simpler, but I can work around.

docs/usage.rst
oliverwm1: Sounds good, focused paragraph on `config_to_yaml`. Might be worth adding a separate section on serializing/deserializing fv3config config dicts later on.

tests/test_diag_table.py
```diff
@@ -0,0 +1,203 @@
...
+2000 1 1 0 0 0
+
+"first_diagnostics", 1, "hours", 1, "hours", "time"
+"second_diagnostics", 1, "hours", 1, "hours", "time"
+
+"dynamics", "u850", "zonal_wind_at_850hPa", "first_diagnostics", "all", "none", "none", 2
+"dynamics", "u850", "zonal_wind_at_850hPa", "first_diagnostics", "all", "none", "none", 2
+
+"dynamics", "u850", "zonal_wind_at_850hPa", "second_diagnostics", "all", "none", "none", 2
+"dynamics", "u850", "zonal_wind_at_850hPa", "second_diagnostics", "all", "none", "none", 2
```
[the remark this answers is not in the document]
oliverwm1: Okay I'm going to leave as is for now. Can always add if we think it's useful later. I see these as a low-level repr of the diag_table that is primarily intended for the Fortran model, so comment headers don't seem essential.

PR #126, opened 2021-02-23
Update for nudging on write_run_directory if necessary

oliverwm1:
fv3config should provision the necessary `patch_files` at the time of `write_run_directory` if nudging is activated. This will simplify scripts that use `fv3config` since they won't have to do this themselves.

### DIFF
--- a/fv3config/config/rundir.py
+++ b/fv3config/config/rundir.py
@@ -17,7 +17,7 @@ def write_run_directory(config, target_directory):
         target_directory (str): target directory, will be created if it does not exist
     """
     logger.debug(f"Writing run directory to {target_directory}")
-    if "gfs_analysis_data" in config:
+    if config["namelist"]["fv_core_nml"].get("nudge", False):
         update_config_for_nudging(config)
     write_assets_to_directory(config, target_directory)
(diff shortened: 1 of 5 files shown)


fv3config/config/rundir.py
```diff
@@ -16,6 +17,8 @@ def write_run_directory(config, target_directory):
         target_directory (str): target directory, will be created if it does not exist
     """
     logger.debug(f"Writing run directory to {target_directory}")
+    if "gfs_analysis_data" in config:
```
nbren12, line 20: This seems fine. But would it be better to switch this on based on the value of the `fv_core_nml.nudge` namelist option? Then, the base_yaml could have `gfs_analysis_data` default options without triggering a download for non-nudging runs.

nbren12 (approved):

fv3config/config/rundir.py
oliverwm1: Went back and for on this, yeah I think you're right (switching on `fv_core_nml.nudge` is how the CLI behaves). The `write_run_directory` will fail if `fv_core_nml.nudge` is True and the user does not provide a `gfs_analysis_data` section, which is good.

PR #129, opened 2021-02-23
Don't specify consistency, bump version to v0.6.1

### LINKED ISSUE #124: Objects lacking md5 sum will fail to download
Not all GCS objects are guaranteed to have md5 hashes. In particular, composite objects will lack them: https://cloud.google.com/storage/docs/composite-objects#metadata.

So, using the `consistency='md5'` flag can resulting in failing downloads during `write_run_directory` if one of the assets is a GCS object without an `md5` hash. We should probably use a more relaxed check like `consistency='size'` or just remove it entirely.

https://github.com/VulcanClimateModeling/fv3config/blob/945d87c773f530ddeca9d2af4554cc0d3bc9fc7f/fv3config/filesystem.py#L36

Related: #70 dask/gcsfs#338

oliverwm1:
To allow provisioning of composite GCS objects, remove `consistency` kwarg from the `fsspec.filesystem` call.

Consistency can still be configured for gcsfs by setting environmental variables.

Since this is a small change, I did an fv3config version bump with the PR.

Resolves #124


nbren12 (approved):

nbren12:
I'm not sure why jenkins is failing. It looks like it succeeded? @ofuhrer?

ofuhrer:
Poor Jenkins seems to trip on `validate_http_response` on the GCP instances, see also... https://jenkins.ginko.ch/job/fv3config_PR/slave=gce-cpu/82/console

Does that ring a 🔔? I've relaunched in the meantime just to see if this was a hicups...

image

ofuhrer:
launch jenkins

PR #131, opened 2021-03-05
Use serialization logic more widely

nbren12:
This PR modifies the serialization APIs and expands their usage.

The APIs are changed to work with file-like objects rather than paths, and use the canonical names `dump/load` commonly used in e.g. pickle, yaml, json, etc. Using file like objects is preferred to paths, since it should be possible to test a serialization code without writing to disk. The new implementation allows users to use any file-like object they like, not just the ones provided by fsspec.

Other changes:

- Simplified the serialization tests by making the `DiagTable` a dataclass (with the validation logic in a `__post_init__` method). 
- Removed as many `import yaml`s as possible from the docs and the codebase.
- Moved the `c12_config.yml` file into a python dictionary. This avoids a chicken-or-egg problem with loading the test configurations from a yaml, only to test loading them from a yaml later on. Put another way, this allows us to avoid calling serialization logic to test units that don't require serialization. It also allows removing path logic from the test files.
- Requires `toolz`. This libraries has routines to avoid making in place updates to dicts.


tests/test_serialization.py
```diff
@@ -0,0 +1,39 @@
...
+    f = io.StringIO()
+    config = assoc(c12_config, "diag_table", diag_table)
+    fv3config.dump(config, f)
+    f.seek(0)
+    loaded = fv3config.load(f)
+
+    if isinstance(config["diag_table"], fv3config.DiagTable):
+        assert config == loaded
+    else:
+        assert dissoc(config, "diag_table") == dissoc(loaded, "diag_table")
```
oliverwm1, line 39:
```suggestion
    assert config == loaded
```
not sure why we need to remove the `diag_table` if it's not a `fv3config.DiagTable` object. I checked and the test passes with this change.

tests/test_serialization.py
```diff
@@ -0,0 +1,39 @@
+import fv3config
+import io
+import datetime
+import pytest
+from toolz import assoc, dissoc
```
oliverwm1, line 5: would prefer to not add this dep.

oliverwm1 (changes requested):
Generally looks great, these methods will be very useful! Some small changes requested.

I would prefer we not add `toolz` as a dependency just for a single usage in one test... this library has a precedent of manually calling `deepcopy` when necessary, can we just do that for this case too?

nbren12:
Can we leave the toolz dependency? It's lightweight, stable, and used by many libraries. I can refactor some usages of deepcopy in the fv3config if it helps.

 I agree deepcopy can accomplish the same thing, but assoc and dissoc are valuable tools for a library that depends so heavily on dicts. Using deepcopy doesn't do enough IMO to discourage mutating dictionaries in place and is less convenient.
I think the tests nicely demonstrate the advantages of dissoc and assoc:
```
assert dissoc(config, "diag_table") == dissoc(loaded, "diag_table")
```
vs
```
left = copy.deepcopy(config)
right = copy.deepcopy(loaded)

del left['diag_table']
del right['diag_table']

assert left == right
```
It is much easier to make a bug for the latter block of code.

tests/test_serialization.py
nbren12: https://github.com/VulcanClimateModeling/fv3config/pull/131#issuecomment-791649574

docs/usage.rst
```diff
@@ -182,9 +182,8 @@ Then a ``DiagTable`` object can be initialized as:
 .. code-block:: python
 
     >>> import fv3config
-    >>> import yaml
     >>> with open("sample_diag_table.yaml") as f:
-            diag_table_dict = yaml.safe_load(f)
+            diag_table_dict = fv3config.load(f)
```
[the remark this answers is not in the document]
nbren12: Thanks for catching this. It would be nice to actually run the code in these examples.  I noticed another place where they contain bugs.

fv3config/config/_serialization.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Args:
+        config: an fv3config object
+        f: the file like object to write to
+
+    """
+    config_copy = deepcopy(config)
+    if isinstance(config["diag_table"], DiagTable):
+        config_copy["diag_table"] = config["diag_table"].asdict()
+    yaml.dump(config_copy, f)
```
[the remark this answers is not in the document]
nbren12: Can you clarify your suggestion? Your code suggestion makes sense to me, but I don't see how the comment relates.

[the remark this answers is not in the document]
nbren12: I guess safe_dump will fail if it encounters an exotic python object?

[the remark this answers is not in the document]
oliverwm1: > I guess safe_dump will fail if it encounters an exotic python object?

that's correct

tests/test_serialization.py
nbren12: Good catch. I think this was cruft from an earlier stage in the PR when I didn't understand why the tests were actually failing.

tests/test_serialization.py
nbren12: removed

docs/usage.rst
```diff
@@ -229,7 +229,7 @@ String representations of the ``diag_table`` (i.e. those expected by the Fortran
 with the :py:meth:`fv3config.DiagTable.from_str` method.
 
 If serializing an ``fv3config`` configuration object to yaml it is recommended to use
-:py:meth:`fv3config.config_to_yaml`. This method will convert any ``DiagTable`` instances to
+:py:meth:`fv3config.load`. This method will convert any ``DiagTable`` instances to
```
oliverwm1, line 232:
```suggestion
:py:meth:`fv3config.dump`. This method will convert any ``DiagTable`` instances to
```

oliverwm1 (approved):
Approving, but found one more issue in the docs.

PR #135, opened 2021-03-30
Ignore orography if empty list

ofuhrer:
This PR ignores orography, in case the user specifies an empty list. There's probably a better way to do this, so I am a taker for alternatives.

### DIFF
--- a/fv3config/_asset_list.py
+++ b/fv3config/_asset_list.py
@@ -38,12 +38,13 @@ def ensure_is_list(asset):
 
 def get_orographic_forcing_asset_list(config):
     """Return asset_list for orographic forcing"""
-    source_directory = get_orographic_forcing_directory(config)
-    if not len(source_directory):
-        return []
-    return asset_list_from_path(
-        source_directory, target_location="INPUT", copy_method="link"
-    )
+    if is_dict_or_list(config["orographic_forcing"]):
+        return ensure_is_list(config["orographic_forcing"])
+    else:
+        source_directory = get_orographic_forcing_directory(config)
+        return asset_list_from_path(
+            source_directory, target_location="INPUT", copy_method="link"
+        )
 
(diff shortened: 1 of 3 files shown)


fv3config/_asset_list.py
```diff
@@ -39,6 +39,8 @@ def ensure_is_list(asset):
 def get_orographic_forcing_asset_list(config):
     """Return asset_list for orographic forcing"""
     source_directory = get_orographic_forcing_directory(config)
+    if not len(source_directory):
+        return []
```
mcgibbon, line 43: Instead of writing special case code for an empty list, you can re-write this function to have the same form as `get_base_forcing_asset_list`, but using `get_orographic_forcing_directory` and using the different arguments for `asset_list_from_path`. This will enable passing an "asset list", from which all entries will be copied. If you pass an empty list, it copies no files.

fv3config/_datastore.py
```diff
@@ -51,6 +51,8 @@ def get_orographic_forcing_directory(config):
     if "orographic_forcing" not in config:
         raise ConfigError("config dictionary must have an 'orographic_forcing' key")
     parent_dirname = config["orographic_forcing"]
+    if not len(parent_dirname):
+        return ""
```
mcgibbon, line 55: If you make the change above, you won't need this code anymore because this should only ever be a string if this function is called.

```suggestion
```

mcgibbon (changes requested):

mcgibbon:
Code looks great! You've added the functionality that orographic forcing can be passed as an asset list instead of only as a directory path. Can you look in the tests to see what test cases exist for `initial_conditions` for asset lists and add the same tests for `orographic_forcing`?

mcgibbon (approved):
LGTM.

PR #137, opened 2021-04-22
Ensure write_run_directory does not mutate input config

### LINKED ISSUE #130: write_run_directory mutates the config dict
I noticed this when trying to memoize a function that called write_run_directory. 

@mcgibbon

oliverwm1:
Currently the `write_run_directory` function mutates the input if nudging is enabled. It also does so (in a subtle way) if the `config["initial_conditions"]` entry is a list. This PR adds tests for these cases and fixes them.

Resolves #130

### DIFF
--- a/fv3config/_asset_list.py
+++ b/fv3config/_asset_list.py
@@ -276,7 +276,8 @@ def check_asset_has_required_keys(asset):
 def config_to_asset_list(config):
     """Convert a configuration dictionary to an asset list. The asset list
     will contain all files for the run directory except the namelist."""
-    asset_list = get_initial_conditions_asset_list(config)
+    asset_list = []
+    asset_list += get_initial_conditions_asset_list(config)
     asset_list += get_base_forcing_asset_list(config)
     asset_list += get_orographic_forcing_asset_list(config)
     asset_list.append(get_field_table_asset(config))
--- a/fv3config/config/nudging.py
+++ b/fv3config/config/nudging.py
@@ -1,4 +1,5 @@
 import collections
+from copy import deepcopy
 from datetime import datetime, timedelta
 import os
 from typing import Sequence, List, Mapping
@@ -119,8 +120,8 @@ def _clear_nudging_assets(config):
         config["patch_files"] = _non_nudging_assets(config["patch_files"], pattern)
 
 
-def update_config_for_nudging(config: Mapping):
-    """Update config object in place to include nudging file assets and associated
+def update_config_for_nudging(config: Mapping) -> Mapping:
+    """Return config object with necessary nudging file assets and associated
     file_names namelist entry. Requires 'gfs_analysis_data' entry in fv3config object
     with 'url' and 'filename_pattern' entries.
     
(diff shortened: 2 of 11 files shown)


fv3config/_asset_list.py
```diff
@@ -276,7 +276,8 @@ def check_asset_has_required_keys(asset):
 def config_to_asset_list(config):
     """Convert a configuration dictionary to an asset list. The asset list
     will contain all files for the run directory except the namelist."""
-    asset_list = get_initial_conditions_asset_list(config)
```
oliverwm1, line 279: for certain conditions, this expression resolves to `asset_list = config["initial_conditions"]` in which case this entry in the config dict gets modified by later modifications of the `asset_list`

tests/test_write_run_directory.py
```diff
@@ -155,3 +155,35 @@ def test_rundir_contains_nudging_asset_if_enabled(self):
...
+            fv3config.get_asset_dict(
+                "gs://vcm-fv3config/data/initial_conditions/gfs_c12_example/v1.0",
+                "initial_conditions_file",
+                target_location="explicit",
+            )
+        ]
+        config_copy = copy.deepcopy(config)
+        with tempfile.TemporaryDirectory() as rundir:
+            fv3config.write_run_directory(config, rundir)
+            assert config == config_copy
```
nbren12, line 189: There is a lot of copy paste in these tests. Can you make this a pytest with `parametrize`?

nbren12 (changes requested):
Thanks for fixing this confusing behavior.

- Please update the changelog to note that this is a breaking change
- I'd also request that we use more modern patterns in these new tests now that pytest is a dependency of this project.

fv3config/config/nudging.py
```diff
@@ -119,8 +120,8 @@ def _clear_nudging_assets(config):
         config["patch_files"] = _non_nudging_assets(config["patch_files"], pattern)
 
 
-def update_config_for_nudging(config: Mapping):
-    """Update config object in place to include nudging file assets and associated
+def update_config_for_nudging(config: Mapping) -> Mapping:
```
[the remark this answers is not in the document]
oliverwm1: It is public, so listed this as a breaking change. I renamed it to `enable_nudging` which is consistent with the already-existing `enable_restart` function which also returns a new dict.

I also renamed the command line entrypoint for this function, which is also a breaking change.

tests/test_write_run_directory.py
```diff
@@ -155,3 +155,35 @@ def test_rundir_contains_nudging_asset_if_enabled(self):
             fv3config.write_run_directory(config, rundir)
             assert "20160801_00.nc" in os.listdir(os.path.join(rundir, "INPUT"))
             assert "20160801_06.nc" in os.listdir(os.path.join(rundir, "INPUT"))
+
+    def test_write_run_directory_does_not_mutate_config(self):
+        config = copy.deepcopy(DEFAULT_CONFIG)
```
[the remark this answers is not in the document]
oliverwm1: Okay, removed `DEFAULT_CONFIG` usages from the new code and tested using pytest where possible.

tests/test_write_run_directory.py
oliverwm1: Done. Had to define a `update_recursive` function, couldn't figure out how to parametrize cleanly without it.

nbren12 (approved):
Thanks Oli. This is ready to merge.

PR #144, opened 2022-04-07
Fix `diag_table` bug preventing reproducible restarts

### LINKED ISSUE #147: Current `diag_table` writing procedure prevents reproducible restarts
### Introduction

After some debugging, I traced the restart reproducibility issue we were experiencing to how `fv3config` writes diagnostics tables.  When writing a `diag_table`, FV3GFS requires a two-line header.  The first line gives a name to the simulation -- it gets attached as metadata to some output files.  The second line is a date.  How we treat this date has an important influence on how the model behaves, because of the code [here](https://github.com/ai2cm/fv3gfs-fortran/blob/9030cf9c9241d087e8fcdef71ab9458aee980609/FV3/coupler_main.F90#L355-L372):

```fortran
    call diag_manager_init (TIME_INIT=date)

!----- always override initial/base date with diag_manager value -----

    call get_base_date ( date_init(1), date_init(2), date_init(3), &
                         date_init(4), date_init(5), date_init(6)  )

!----- use current date if no base date ------

    if ( date_init(1) == 0 ) date_init = date

!----- set initial and current time types ------

    Time_init  = set_date (date_init(1), date_init(2), date_init(3), &
                           date_init(4), date_init(5), date_init(6))

    Time_atmos = set_date (date(1), date(2), date(3),  &
                           date(4), date(5), date(6))
```

`Time_init` and `Time_atmos` are key variables in initializing the model:
- `Time_init` is intended to represent the time that a simulation was first initialized -- in the case of a segmented run, this is the time of the start of the entire simulation, not the start of segment. 
- `Time_atmos` is the current time of the run -- in the case of a segmented run, this is the time at the start of the segment.  In the case of a segmented run, this is read in from the third line of `coupler.res` restart file.

`Time_init` is set based on the value of `date_init`.  What the code above shows is that `date_init` -- while it can be set earlier in `coupler_main.F90` based on its value in the `coupler.res` restart file -- is always overridden by the date in the diagnostics table (this is what the `get_base_date` subroutine does).  

### What `fv3config` currently does

When writing the `diag_table` in the case of a segmented run, `fv3config` will use the date from the third line of the `coupler.res` restart file (the current date rather than the initial date): https://github.com/ai2cm/fv3config/blob/544eaf1bc6f1c4617cd8ee6bd3298136ed180f4c/fv3config/config/derive.py#L74-L91 

### Why does this lead to irreproducible restarts?

There may be other downstream impacts of this issue, but the first place it shows up is in the [computation of the solar hour](https://github.com/ai2cm/fv3gfs-fortran/blob/9030cf9c9241d087e8fcdef71ab9458aee980609/FV3/gfsphysics/GFS_layer/GFS_driver.F90#L499).  It depends on the amount of time between the initialization time of the simulation and the current time, `Model%phour`.  For segment lengths that evenly divide 24 hours, this may not be a problem; however, it is where short restart reproducibility tests -- which preferably have segment lengths of less than a day -- get tripped up first.  `Model%phour` and the initialization date are also used in the [surface cycling](https://github.com/ai2cm/fv3gfs-fortran/blob/9030cf9c9241d087e8fcdef71ab9458aee980609/FV3/gfsphysics/physics/gcycle.F90#L168-L170), and the initialization time is used to set the value of [a random seed for a subgrid cloud scheme](https://github.com/ai2cm/fv3gfs-fortran/blob/9030cf9c9241d087e8fcdef71ab9458aee980609/FV3/gfsphysics/physics/rad_initialize.f#L186), which could be other sources of irreproducibility in segmented runs.

Basically the model expects the initialization time to stay constant, but we are changing it in every segment of segmented runs.

### Proposed solution

The natural solution, which requires no changes to the fortran model, is to address this in `fv3config` by using the date from the second line of `coupler.res` instead of the third line when writing the `diag_table`.  This way the simulation's initialization date is always propagated through the `coupler.res` restart files until the end of the run.  The initialization date would then be constant for the `diag_manager`, which I think is more in line with what it expects as well.

For reference, GFDL runscripts manage this by having the initialization date defined as a constant, which is always used when writing the `diag_table` to the run directory in subsequent segments.  

#### Implications

- Run segment lengths must now evenly divide the radiation timesteps.  The radiative transfer code must run during the first timestep of a segment to initialize some variables.  If it does not, a segmentation fault will occur in the physics.  Previously, the radiative transfer code would always run on the first timestep of a segment, regardless of whether it meant that the radiation would be called sooner than it would have been in a continuous run (this was another side of effect of always setting the initialization time to the start of the segment).  
- We likely will want to handle configuring to start from coarsened restart files in a specialized manner.  This is different than a segmented run in that we *do not* necessarily want to set the initialization date to what it is in the `coupler.res` file (because this initialization time corresponds to what it was in the fine-resolution run).  If we set it to what it was in the `coupler.res` file we would be limited to starting the model only from certain timesteps due to the issue above; the solution in this case would be to do what we do now, i.e. set the date in `diag_table` to the time associated with the set of restart files -- but only when first initializing the run -- and then follow the reproducible restart pathway for the rest of the simulation.

spencerkclark:
This PR introduces a `get_time_configuration` function in place of the existing `get_current_date` function.  The `get_time_configuration` function returns both the initialization date and the current date.

The initialization and current dates can be found in lines two and three of any `coupler.res` restart file, and are read in accordingly when one is present in a set of restart files provided as an initial condition.  If the initial condition does not contain a `coupler.res` file, or `coupler_nml.force_date_from_namelist` is set to true, the initialization date and current date are set to the value of the `coupler_nml.current_date` namelist parameter.  

As found in #147, for reproducible restarts, it is critical that the base date written in the diagnostics table is the initialization date of the run (therefore it will remain constant across segments).  Previously in segmented runs, we reset the base date to the date of the current segment, which prevented reproducible restarts.  This PR fixes this issue (see [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2022-04-06-restart-reprodicibility/2022-04-07-reproducible-restart.ipynb)).

Resolves #147
See also: ai2cm/fv3gfs-fortran#269


fv3config/config/derive.py
```diff
@@ -42,33 +42,115 @@ def get_run_duration(config):
...
+    """Return initialization_date specified in coupler.res file
+
+    Args:
+        coupler_res_filename (str): a coupler.res filename
+
+    Returns:
+        list: initialization_date as list of ints [year, month, day, hour, min, sec]
+    """
+    iniitalization_date, _ = _read_coupler_res_dates(coupler_res_filename)
+    return iniitalization_date
```
oliverwm1, line 153:
```suggestion
    initialization_date, _ = _read_coupler_res_dates(coupler_res_filename)
    return initialization_date
```

oliverwm1 (approved):
Thanks for tracking this down @spencerkclark! I think your fix looks good. I agree this is not a big issue for runs initialized from coarsened restarts, since we typically don't even have a `coupler.res` file for those and we set the initialization date through the namelist.

nbren12 (changes requested):
Can you add a note to the history.md describing in as few words as possible what user facing change is? I'm having trouble parsing this from the PR description thanks.

fv3config/config/derive.py
```diff
@@ -42,33 +42,115 @@ def get_run_duration(config):
...
+    existence of a coupler.res file, and if it exists, returns the initialization_date
+    written within it. If a coupler.res files does not exist, the current_date
+    specified in the namelist or a null date is used.  To FV3GFS, a null date and
+    the current date are equivalent.
+
+    Returning the initialization_date instead of the current_date if a coupler.res
+    file exists is important for obtaining reproducible restarts in segmented runs.
+    Another advantage is that it harmonizes the units used to encode times in
+    fotran-generated diagnostics files across segments. See more discussion in
+    GH 147.
```
nbren12, line 103: I find this docstring pretty hard to follow would be easier to just see the if-statements in the code.

fv3config/config/derive.py
```diff
@@ -42,33 +42,115 @@ def get_run_duration(config):
...
+
+    Args:
+        config (dict): a configuration dictionary
+
+    Returns:
+        list: base_date as list of ints [year, month, day, hour, min, sec]
+    """
+    return _get_date(config, _get_initialization_date_from_coupler_res)
```
nbren12, line 111: Is this `_get_date` indirection needed? Pretty hard to tell what's going on.

fv3config/config/derive.py
```diff
@@ -42,33 +42,115 @@ def get_run_duration(config):
     )
 
 
-def get_current_date(config):
-    """Return current_date from configuration dictionary. This function may read from
+def _get_date(config, coupler_res_date_getter):
+    """Return date from configuration dictionary.  This function may read from
     the remote initial_conditions path in the given configuration dictionary.
```
nbren12, line 47: > This function may read from
    the remote initial_conditions path in the given configuration dictionary.
 
Maybe doing I/O is pretty confusing...can you pull the I/O up the call-stack?

fv3config/config/derive.py
```diff
@@ -42,33 +42,115 @@ def get_run_duration(config):
...
+
+def _read_coupler_res_dates(coupler_res_filename):
+    """Read the dates contained in a coupler.res file
+    
+    Args:
+        coupler_res_filename (str): a coupler.res filename
+
+    Returns:
+        initialization_date (list): date as list of ints [year, month, day, hour, min, sec]
+        current_date (list): date as list of ints [year, month, day, hour, min, sec]
```
nbren12, line 133: Lot's of redundant text here that could be expressed in types.

fv3config/config/derive.py
```diff
@@ -42,33 +42,115 @@ def get_run_duration(config):
...
+            f"(line must contain six integers)"
+        )
+    return date
+
+
+def _read_coupler_res_dates(coupler_res_filename):
+    """Read the dates contained in a coupler.res file
+    
+    Args:
+        coupler_res_filename (str): a coupler.res filename
```
nbren12, line 129: This unneeded IMO
```suggestion
```

nbren12 (changes requested):
I'm having some trouble following this code. Do you have some time to VC about it? Some things that would help:

1. use datetime.datetime rather than lists of ints in all the low level helpers
2. Removing all the helpers. I can't follow all the indirection in the code. The `_get_date` function buries I/O and conditional logic in a low-level helper. Maybe just copy paste all of this as if-statements in `fv3config/config/rundir.py`?

fv3config/config/derive.py
nbren12: e.g. by returning a `datetime.datetime` object.

spencerkclark:
> I'm having some trouble following this code. Do you have some time to VC about it?

Sure.  Most of the code/docstrings are adapted from what existed before, but I can rewrite it in a different way.

tests/test_tables.py
```diff
@@ -163,6 +166,24 @@ def test_get_current_date_from_bad_coupler_res(self):
...
+    def test_get_initialization_date_from_bad_coupler_res(self):
+        rundir = self.make_run_directory("test_rundir")
+        coupler_res_filename = os.path.join(rundir, "coupler.res")
+        with open(coupler_res_filename, "w") as f:
+            f.write(bad_coupler_res)
+        with self.assertRaises(ConfigError):
+            _get_initialization_date_from_coupler_res(coupler_res_filename)
+
     def test_get_current_date_from_config_force_date_true(self):
         config = copy.deepcopy(DEFAULT_CONFIG)
```
nbren12, line 188:
```suggestion
        config = c12_config()
```

tests/test_tables.py
```diff
@@ -192,6 +213,37 @@ def test_get_current_date_from_config_which_includes_coupler_res_asset(self):
         current_date = get_current_date(config)
         self.assertEqual(current_date, valid_current_date)
 
+    def test_get_diag_table_base_date_from_config_force_date_true(self):
+        config = copy.deepcopy(DEFAULT_CONFIG)
```
nbren12, line 217:
```suggestion
        config = c12_config()
```

tests/test_tables.py
```diff
@@ -192,6 +213,37 @@ def test_get_current_date_from_config_which_includes_coupler_res_asset(self):
...
+        config["namelist"]["coupler_nml"]["force_date_from_namelist"] = True
+        config["namelist"]["coupler_nml"]["current_date"] = valid_current_date
+        base_date = get_diag_table_base_date(config)
+        self.assertEqual(base_date, valid_current_date)
+
+    def test_get_diag_table_base_date_from_config_force_date_false(self):
+        config = copy.deepcopy(DEFAULT_CONFIG)
```
nbren12, line 224:
```suggestion
        config = c12_config()
```

tests/test_tables.py
```diff
@@ -192,6 +213,37 @@ def test_get_current_date_from_config_which_includes_coupler_res_asset(self):
...
+        config["namelist"]["coupler_nml"]["force_date_from_namelist"] = False
+        config["namelist"]["coupler_nml"]["current_date"] = valid_current_date
+        base_date = get_diag_table_base_date(config)
+        self.assertEqual(base_date, valid_current_date)
+
+    def test_get_diag_table_base_date_from_config_which_includes_coupler_res_asset(
+        self,
+    ):
+        config = copy.deepcopy(DEFAULT_CONFIG)
```
nbren12, line 233:
```suggestion
        config = c12_config()
```

fv3config/config/derive.py
nbren12: this is an optional change.

tests/test_tables.py
spencerkclark: I updated all uses in this file.

spencerkclark: I updated all uses in this file.

fv3config/config/derive.py
spencerkclark: Thanks -- I will hold off on this for now.  While I'm not sure if it is possible to run FV3GFS with a calendar other than "julian" (or whether we care to support that) the `coupler.res` file does include calendar type information, which we might make use of in constructing dates:
```
    2        (Calendar: no_calendar=0, thirty_day_months=1, julian=2, gregorian=3, noleap=4)
  2016     8     1     0     0     0        Model start time:   year, month, day, hour, minute, second
  2016     8     3     0     0     0        Current model time: year, month, day, hour, minute, second
```
For the time being I think sticking with the calendar-agnostic list that was used previously is fine.

spencerkclark: Thanks -- I will hold off on this for now.  While I'm not sure if it is possible to run FV3GFS with a calendar other than "julian" (or whether we care to support that) the `coupler.res` file does include calendar type information, which we might make use of in constructing dates:
```
    2        (Calendar: no_calendar=0, thirty_day_months=1, julian=2, gregorian=3, noleap=4)
  2016     8     1     0     0     0        Model start time:   year, month, day, hour, minute, second
  2016     8     3     0     0     0        Current model time: year, month, day, hour, minute, second
```
For the time being I think sticking with the calendar-agnostic list that was used previously is fine.

spencerkclark:
Thanks for chatting yesterday @nbren12.  I think I addressed everything we discussed.

> Food for later thought, it would be nice to load the information in coupler.res into the config at an earlier stage so that I/O is not needed to figure out the date from a loaded config.

Yes, this would be nice.  The `coupler_nml.current_date` parameter might actually already encode the base date information.  I think it needs to be set on initialization, but is left untouched in subsequent segments, and so would be propagated forward with the config.  It might be a little confusing to rely on that assumption though.

nbren12 (approved):
Thanks for the changes. Very easy to follow now. Just one small comment above.

fv3config/config/derive.py
```diff
@@ -42,53 +42,57 @@ def get_run_duration(config):
...
+    This function may read from the remote initial_conditions path given in
+    the configuration dicitionary.
 
     Args:
         config (dict): a configuration dictionary
 
     Returns:
-        list: current_date as list of ints [year, month, day, hour, min, sec]
+        initialization_date (list): date as list of ints [year, month, day, hour, min, sec]
+        current_date (list): date as list of ints [year, month, day, hour, min, sec]
```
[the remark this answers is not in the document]
spencerkclark: Good idea.  Should hopefully be a little clearer now.

PR #152, opened 2022-04-14
Fix license classifier in `setup.py`

spencerkclark:
PyPI requires classifiers come from [this list](https://pypi.org/classifiers/).  This appears to be the appropriate one for Apache-licensed tools.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
     classifiers=[
         "Development Status :: 2 - Pre-Alpha",
         "Intended Audience :: Developers",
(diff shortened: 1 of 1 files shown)


nbren12 (approved):
Ah. twine/pypi is pretty picky. maybe you can use test pypi before merging prs to make sure it works: https://packaging.python.org/en/latest/guides/using-testpypi/

spencerkclark:
Thanks, yeah, I was a bit overconfident in this case...I confirmed with this change it works now.

PR #153, opened 2022-04-15
Make fv3config.config_to_asset_list public API

nbren12:
I would like to enable wandb (and other metadata tracking) integrations for assets supported by fv3config. The "asset list" is the natural target for this, but is not currently a complete representation of the config dict since there is lingering logic in the `write_run_directory` function.

* push all config-handling logic into `config_to_asset_list`
* make `config_to_asset_list` public API

This will allow usages like this downstream:
```
assets = fv3config.config_to_asset_list(config)
for asset in assets:
    if asset["source_location"].startswith("wandb://"):
        artifact = run.use_artifact(...asset)
        path = artifact.download()
        asset["source_location"] = path
    fv3config.write_asset(asset, "rundir")
```

To reviewers: I recommend reviewing commit-by-commit. I tried to isolate the refactors into coherent commits, and it should be much easier to follow that way.

### DIFF
--- a/fv3config/_asset_list.py
+++ b/fv3config/_asset_list.py
@@ -1,25 +1,15 @@
-import io
+"""Assets
+
+Assets represent either remote, local, or in memory data that can be written to
+a local disk
+"""
 import logging
 import os
 
-from ._datastore import (
-    get_initial_conditions_directory,
-    get_orographic_forcing_directory,
-    get_base_forcing_directory,
-)
-from fv3config._datastore import (
-    get_field_table_filename,
-    get_diag_table_filename,
-    get_data_table_filename,
-)
-from .config.diag_table import DiagTable
-from .config.namelist import config_to_namelist
-from .config._serialization import dump
 from ._exceptions import ConfigError
 from . import filesystem
 
 
-FV3CONFIG_YML_NAME = "fv3config.yml"
 logger = logging.getLogger("fv3config")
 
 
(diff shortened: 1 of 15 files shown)


fv3config/_asset_list.py
```diff
@@ -231,6 +165,8 @@ def write_asset(asset, target_directory):
         logger.debug(f"Writing asset bytes to {target_path}.")
         with open(target_path, "wb") as f:
             f.write(asset["bytes"])
+    elif "directory" in asset:
```
spencerkclark, line 168: It seems more natural to check the value stored under the `"directory"` key than to check whether the `"directory"` key exists in the asset dictionary:
```suggestion
    elif asset.get("directory", False):
```

fv3config/_asset_list.py
```diff
@@ -231,6 +165,8 @@ def write_asset(asset, target_directory):
...
+        return os.makedirs(target_path, exist_ok=True)
     else:
         raise ConfigError(
             "Cannot write asset. Asset must have either a `copy_method` or `bytes` key."
```
spencerkclark, line 172: Maybe update this message to something like: "Cannot write asset.  Asset must be a directory or have either a \`copy_method\` or \`bytes\` key."

spencerkclark (approved):
Nice clean refactor.  At some point it might be nice to formalize what an asset is in the form of an object (rather than a dictionary, to which one can add arbitrary new keys), but I think that's outside the scope of this PR.

fv3config/_asset_list.py
nbren12: i changed to making `directory` a `copy_method`.

nbren12:
@spencerkclark It would be nice to switch to use types throughout fv3config.

